PR #65, opened 2022-10-05
Small fix to remove runtime errors from NA's

qubixes:
It also resolves some more problems with mixed integer columns, which apparently have been recently introduced into pandas. Adds tests to prevent the same issue in the future (although runtime errors won't be caught).


metasynth/distribution/base.py
```diff
@@ -35,6 +35,8 @@ def fit(cls, series: Sequence, *args, **kwargs) -> BaseDistribution:
             Fitted distribution.
         """
         pd_series = cls._to_series(series)
+        if len(pd_series) == 0:
+            return cls._example_distribution()
```
vankesteren, line 39: Ok nice; 

I notice we are using example distribution a lot throughout, should we call it something different? Default distribution?

vankesteren (approved):

vankesteren:
Approved! check my comment above but that could be done in a different PR; we can convert it to a very small issue.

metasynth/distribution/base.py
qubixes: > Ok nice;
> 
> I notice we are using example distribution a lot throughout, should we call it something different? Default distribution?

Yes, I wasn't planning to use it that much, but now that it's usage has increased, maybe a `default_distribution` makes more sense. Let's do it in another pull request indeed.

PR #66, opened 2022-10-05
Possible polars version of MetaSynth

qubixes:
It seems to mostly work, with some part still to be implemented. Notably:

- dtype stuff, need to figure out polars dtypes from strings.
- categories: Currently untested/not used in the example code.
- Tests: Most are not working.


metasynth/distribution/base.py
```diff
@@ -203,7 +204,7 @@ def __getattr__(self, attr: str):
 
     @classmethod
     def _fit(cls, values):
-        param = cls.dist_class.fit(values[~np.isnan(values)])
+        param = cls.dist_class.fit(values.drop_nans())
```
qubixes, line 207: Change to drop_nulls

vankesteren:
So decision on this: We will continue exploring using polars as backend because it streamlines our code.

Important task: find out how well pandas <-> polars conversion works!

PS: I also just found the [tidypolars](https://tidypolars.readthedocs.io/en/latest/index.html) package, which I love already.

metasynth/distribution/base.py
```diff
@@ -42,17 +43,19 @@ def fit(cls, series: Sequence, *args, **kwargs) -> BaseDistribution:
 
     @staticmethod
     def _to_series(values: Sequence):
-        if isinstance(values, pandas.Series):
-            series = values.dropna()
+        if isinstance(values, pl.Series):
+            series = values.drop_nulls()
+        elif isinstance(values, pd.Series):
+            series = pl.Series(values).drop_nulls()  # pylint: disable=assignment-from-no-return
```
vankesteren, line 49: So this is where the conversion happens. This was extensively tested for different data types yeah?

metasynth/distribution/categorical.py
```diff
@@ -1,11 +1,11 @@
 """Module containing categorical distributions."""
 
-from typing import Sequence, Union
+from typing import Union
 
 import pandas as pd
```
vankesteren, line 5: dependence on pandas should be removed here (also in line 58 of this file)

metasynth/distribution/discrete.py
```diff
@@ -116,21 +109,20 @@ def draw(self):
...
-        vals = values[~np.isnan(values)]
-        if np.min(values) < self.low:
+    def _information_criterion(self, values):
+        if values.min() < self.low:
             return 3+999*len(values)
 
         # If the values are not unique the fit is extremely bad.
-        if len(set(vals)) != len(vals):
+        if len(set(values)) != len(values):
             return 3+999*len(values)
```
vankesteren, line 118: This somehow seems sketchy. Do we still use this?

metasynth/distribution/discrete.py
```diff
@@ -116,21 +109,20 @@ def draw(self):
...
-        high = np.max(vals)+1
+        low = values.min()
+        high = values.max()+1
 
         if self.consecutive == 1:
             # Check if the values are truly
-            if len(vals) == high-low and np.all(vals == np.arange(low, high)):
+            if len(values) == high-low and np.all(values.to_numpy() == np.arange(low, high)):
                 return 3
             return 3+999*len(values)
```
vankesteren, line 127: same here

metasynth/var.py
```diff
@@ -70,7 +82,7 @@ def __init__(self,  # pylint: disable=too-many-arguments
                              " prop_missing is None.")
 
     @classmethod
-    def detect(cls, series_or_dataframe: Union[pd.Series, pd.DataFrame],
+    def detect(cls, series_or_dataframe: Union[pd.Series, pl.Series, pl.Series, pl.DataFrame],
```
vankesteren, line 85:
```suggestion
    def detect(cls, series_or_dataframe: Union[pd.Series, pl.Series, pl.DataFrame],
```

metasynth/var.py
```diff
@@ -92,28 +104,32 @@ def detect(cls, series_or_dataframe: Union[pd.Series, pd.DataFrame],
...
+            "int": "discrete",
+            "float": "continuous",
             "date": "date",
-            "datetime64": "datetime",
             "datetime": "datetime",
             "time": "time",
+            "str": "string",
+            "categorical": "categorical"
         }
         try:
```
vankesteren, line 134: I would expect that we don't use this anymore? this should be done by pandas -> polars conversion function and then we only have to maintain one vartype conversion list (namely polars to metasynth).

setup.py
```diff
@@ -23,6 +23,8 @@
     python_requires='~=3.7',
     install_requires=[
         "pandas",
```
vankesteren, line 25: can we somehow make pandas an optional dependency? is there a way to do that in python? as in: package works completely without pandas, but if you want to use the pandas functionality you can install it.

tests/test_dataset.py
```diff
@@ -3,32 +3,46 @@
...
+    "Fare": "float",
+}
+
+
+def _read_csv(fp, dataframe_lib):
+    if dataframe_lib == "pandas":
+        df = pd.read_csv(fp, dtype=dtypes)
+        return df.iloc[:100]
+    else:
+        df = pl.read_csv(fp, dtypes={x: pl.Categorical for x, x_type in dtypes.items() if x_type == "category"})
```
vankesteren, line 35: haha nice

vankesteren (approved):
Conditional approve: can you answer / look at my comments before merging?

vankesteren:
oh yes, could #78 be merged before merging this one?

metasynth/dataset.py
```diff
@@ -115,7 +115,7 @@ def from_dataframe(cls,
             spec = deepcopy(spec)
 
         all_vars = []
-        for col_name in list(df):
+        for col_name in df.columns:
```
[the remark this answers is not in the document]
qubixes: I agree!

metasynth/distribution/base.py
qubixes: It should work for most at least arrays and pandas DataFrames, probably lists as well.

metasynth/distribution/categorical.py
qubixes: Are you sure? I think for the annotations, pandas dataframes should be allowed here.

metasynth/distribution/discrete.py
qubixes: Which part do you find sketchy? I do think this is still used to determine whether we need to send a warning.

metasynth/var.py
qubixes: This is the polars -> metasynth conversion. :) I updated the names/code.

setup.py
qubixes: Yes, but we need pandas for some checks, so we need to set it as a requirement, unless we rearrange the code.

PR #71, opened 2022-10-18
Raise an error when frac_used is outside [0, 1]

### LINKED ISSUE #70: RegexDistribution: give warning as frac_used > 1
bla_distribution = RegexDistribution([(r"[ABCDEF]",1),(r"[0-9]",3)])

qubixes:
Fixes #70

### DIFF
--- a/metasynth/distribution/regex/element.py
+++ b/metasynth/distribution/regex/element.py
@@ -19,6 +19,12 @@ class BaseRegexElement(ABC):
 
     frac_used = 1.0
 
+    def __init__(self, frac_used: float):
+        if frac_used > 1 + 1e-6 or frac_used < -1e-6:
+            raise ValueError(f"Error initializing RegexElement with fraction used {frac_used}."
+                             f"Set the fraction to be used between 0 (never) and 1 (always).")
+        self.frac_used = frac_used
+
     @abstractmethod
     def information_budget(self, regex_stat: Dict) -> float:
         """Addition to the AIC for the current regex."""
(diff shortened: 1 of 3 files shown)


metasynth/distribution/regex/element.py
```diff
@@ -346,10 +358,17 @@ class AnyRegex(BaseRegexClass):
...
+        super().__init__(min_digit, max_digit, frac_used)
+
+        # if frac_used > 1 + 1e-6 or frac_used < -1e-6:
+        #     raise ValueError(f"Error initializing RegexElement with fraction used {frac_used}."
+        #                      f"Set the fraction to be used between 0 (never) and 1 (always).")
+        # self.min_digit = min_digit
+        # self.max_digit = max_digit
+        # self.frac_used = frac_used
+
+    def _prepare_regex(self):
```
vankesteren, line 371: Why is this part commented out?

Also, why the `frac_used < -1e-6` and not just `frac_used < 0.0`? Is there a precision error somewhere?

vankesteren (changes requested):
See my comment!

once fixed, consider this approved :)

metasynth/distribution/regex/element.py
qubixes: I commented it out this part, because it is done in the `super().__init__` part. I gave it a margin, because of possible rounding errors down the line somewhere. I don't think there's any of that happening in our code, but I was worried it would come from user code somewhere.

I will remove the commented out code.

PR #73, opened 2022-10-18
Rename all example datasets to default dataset

### LINKED ISSUE #67: Replace _example_distribution with default_distribution

qubixes:
Also makes it a public method.

Fixes #67


vankesteren:
Yeah, makes sense to make it a public method as well.

PR #77, opened 2022-10-24
Convert integers to nullable-integers if p(NA)>0

### LINKED ISSUE #74: int64 with prop_missing > 0 gives error on empty values when synthesizing
When calling synthesize(), var of type int64 with prop_missing > 0 (details below) sometimes gives:

TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'

presumably on the missing fields (doesn't happen with prop_missing=0).

{ "name": "Last page", "type": "discrete", "dtype": "int64", "prop_missing": 0.1, "distribution": { "name": "DiscreteUniformDistribution", "parameters": { "low": 6, "high": 7 } } },

qubixes:
Fixes #74 

The problem is that pandas makes a distinction between integers and nullable integers. Therefore, a column/series that wasn't originally a nullable integer, needs to be converted to one if prop_missing is manually set.

### DIFF
--- a/metasynth/var.py
+++ b/metasynth/var.py
@@ -69,6 +69,9 @@ class MetaVar():
             raise ValueError(f"Error while initializing variable {self.name}."
                              " prop_missing is None.")
 
+        if self.prop_missing > 0 and str(self.dtype).startswith("int"):
+            self.dtype = pd.Int64Dtype()
+
     @classmethod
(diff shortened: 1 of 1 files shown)


qubixes:
Not necessary anymore since we're using polars internally right now. Please reopen if it is still a problem.

PR #87, opened 2023-02-08
Fix not passing the right distribution tree

qubixes:


vankesteren (commented):
Seems fine, isn't this part of a larger commit? Did the argument name of the .fit method for var change?

qubixes:
> Seems fine, isn't this part of a larger commit? Did the argument name of the .fit method for var change?

This is the whole commit. It is a small bugfix, with no other implications. I found out that in some cases the wrong distribution tree would be used (it defaults to "builtin") while I was working on the metasynth disclosure control repository.

PR #88, opened 2023-02-09
Add test utilities for plugins to check their code

qubixes:
This adds functionality for future extension to quickly test their distributions and see if everything (seemingly) works.

### DIFF
--- a/metasynth/testutils.py
+++ b/metasynth/testutils.py
@@ -1,0 +1,46 @@
+"""Testing utilities for plugins."""
+
+from typing import Optional
+
+import polars as pl
+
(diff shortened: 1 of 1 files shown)


vankesteren (approved):
Oh this is a great idea. Is the plan to use this in the test suite in the plugin repositories? So that you can just run metasynth.testutils.check_dist_type() there?

qubixes:
> Oh this is a great idea. Is the plan to use this in the test suite in the plugin repositories? So that you can just run metasynth.testutils.check_dist_type() there?

Yes, there is already a PR in the metasynth-disclosure repository!

PR #89, opened 2023-02-09
Add pyproject.toml and remove versioneer

### LINKED ISSUE #85: Switch to pyproject.toml

qubixes:
Python 3.7 support will be removed, so that is why the tests are failing. Fixes #85.


pyproject.toml
```diff
@@ -0,0 +1,61 @@
...
+	"pandas",
+	"polars>=0.14.17",
+	"numpy>=1.20",
+    "pyarrow",  # Dependency of polars since we're converting from pandas.
+    "scipy",
+    "numpy>=1.20",
+    "faker",
+    "scikit-learn",
+    "jsonschema",
+    "importlib-metadata;python_version<'3.10'",
```
vankesteren, line 29: Does this mean we need to change this when the python version goes above 3.9?

vankesteren (approved):
Wow this is a lot cleaner than before! Nice work

pyproject.toml
qubixes: Yes, when Python 3.9 becomes unsupported, we can remove the line.

PR #100, opened 2023-06-01
change tutorial to getting started

vankesteren:
Also make it shorter and refer to advanced tutorial


qubixes (changes requested):
I think it makes more sense for the GMF file not to be there when you run the notebook, especially since we tell the user to open the file after running a cell in the notebook.

qubixes (approved):

PR #109, opened 2023-07-04
Remove inheritance in favor of properties

### LINKED ISSUE #95: Idea: remove some inheritance in favor of decorators
One thing that I feel is not quite satisfactory is the multiple inheritance structure in some places of the code. I have read some articles, and perhaps we can do it better with decorators.

```python
def distribution(implements=None, provenance=None, var_type=None, is_unique=None):
    def _wrap(cls):
        if implements is not None:
            cls.implements = implements
#        if provenance is not None:
#            cls.provenance = provenance
        if var_type is not None:
            cls.var_type = var_type
        if is_unique is not None:
            cls.is_unique = is_unique
        return cls
    return _wrap
```

This would be the decorator that you would put in front of the new implementation:

```python
import metasynth as ms

@ms.distribution(implements="core.uniform", var_type="continuous", is_unique=False)
class UniformDistribution(BaseDistribution):
    ...
```

Advantages:

- No multiple inheritance, where the order of the classes matters.
- Can remove the `ContinuousDistribution`, `BaseDisclosure`, etc. classes.
- Possibly easier to understand for developers (perhaps not for beginners?)
- More flexible? I could see a world where the decorator might also register distributions for example.
- Inheritance becomes more optional.

Disadvantages:

- Decorators might be something users don't know about.
- Costs time to implement.

qubixes:
This PR converts some of the inheritance to a property to set class methods. No changes for users. User distributions should be easier to implemented now.

Fixes #95

### DIFF
--- a/metasynth/distribution/__init__.py
+++ b/metasynth/distribution/__init__.py
@@ -5,12 +5,6 @@ somewhat realistic. The concept of distributions here is not only for
 numerical data, but also for generating strings for example.
 """  # pylint: disable=invalid-name
 
-from metasynth.distribution.base import (CategoricalDistribution,
-                                         ContinuousDistribution,
-                                         DateDistribution,
-                                         DateTimeDistribution,
-                                         DiscreteDistribution,
-                                         StringDistribution, TimeDistribution)
 from metasynth.distribution.categorical import MultinoulliDistribution
 from metasynth.distribution.continuous import (ExponentialDistribution,
                                                LogNormalDistribution,
--- a/metasynth/distribution/__init__.py
+++ b/metasynth/distribution/__init__.py
@@ -22,8 +22,6 @@ from metasynth.distribution.regex import (RegexDistribution,
                                           UniqueRegexDistribution)
 
 __all__ = [
-    "DiscreteDistribution", "ContinuousDistribution", "CategoricalDistribution",
-    "DateDistribution", "DateTimeDistribution", "StringDistribution", "TimeDistribution",
     "MultinoulliDistribution", "UniformDistribution", "NormalDistribution",
     "LogNormalDistribution", "TruncatedNormalDistribution", "ExponentialDistribution",
     "DiscreteUniformDistribution", "PoissonDistribution", "UniqueKeyDistribution",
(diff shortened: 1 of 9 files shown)


qubixes:
@Samuwhale It's probably easier if the sphinx documentation is written later, since you're working on it?

metasynth/distribution/__init__.py
```diff
@@ -5,12 +5,6 @@
 numerical data, but also for generating strings for example.
 """  # pylint: disable=invalid-name
 
-from metasynth.distribution.base import (CategoricalDistribution,
-                                         ContinuousDistribution,
-                                         DateDistribution,
-                                         DateTimeDistribution,
-                                         DiscreteDistribution,
-                                         StringDistribution, TimeDistribution)
```
vankesteren, line 13: Ok it's really nice that this can all be removed now

metasynth/distribution/base.py
```diff
@@ -150,53 +149,42 @@ def default_distribution(cls) -> BaseDistribution:
...
-    """Base Class for date distributions."""
-
-    var_type = "date"
-
-
-class TimeDistribution(BaseDistribution):
-    """Base Class for time distributions."""
-
-    var_type = "time"
+def distclass(implements: Optional[str] = None, provenance: Optional[str] = None,
```
vankesteren, line 152: I think the name `distclass` is not descriptive enough here and elsewhere, how about  `metadist` or even `metasynthdistribution`?

So 
```py
@metadist(implements="core.multinoulli", var_type="categorical")
class MultinoulliDistribution(BaseDistribution):
  ...
```

Conceptually, in my mind this decorator makes sure that any distribution that is at least a BaseDistribution interfaces well with the Metasynth package by implementing some additional properties. So it's a metasynth-specific thing, hence `metadist`. But if you have other thoughts feel free to ignore!

vankesteren (approved):
This is a wonderful PR! it's really simplifying the distribution implementations 👍 

Small comment about naming that I would like addressed, but for the rest it looks really good and can be merged.

metasynth/distribution/base.py
qubixes: Let's do metadist, I think that fits the package, and is not too long.

PR #111, opened 2023-07-04
Fix misprop range

### LINKED ISSUE #105: values for `prop_missing` outside (0,1) are accepted, and saved, but silently set to 0 or 1 when applied.
Values for `prop_missing` in the specification that fall outside (0,1) (f.i. `'prop_missing': -10` or `'prop_missing': 100`) are accepted without warning, and saved in the GMF-spec file, yet reset to 0 or 1 when applied.

qubixes:
Ensure that error messages get thrown when the missing proportion is outside [0, 1].

Fixes #105

### DIFF
--- a/metasynth/var.py
+++ b/metasynth/var.py
@@ -72,6 +72,9 @@ class MetaVar():
         if self.prop_missing is None:
             raise ValueError(f"Error while initializing variable {self.name}."
                              " prop_missing is None.")
+        if self.prop_missing < -1e-8 or self.prop_missing > 1+1e-8:
+            raise ValueError(f"Cannot create variable '{self.name}' with proportion missing "
+                             "outside range [0, 1]")
 
     @classmethod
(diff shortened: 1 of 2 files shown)


metasynth/var.py
```diff
@@ -72,6 +72,9 @@ def __init__(self,  # pylint: disable=too-many-arguments
         if self.prop_missing is None:
             raise ValueError(f"Error while initializing variable {self.name}."
                              " prop_missing is None.")
+        if self.prop_missing < -1e-8 or self.prop_missing > 1+1e-8:
```
vankesteren, line 75: That looks like a very weird check. Why not:
```suggestion
        if self.prop_missing < -1.0 or self.prop_missing > 1.0:
```

is this to do with how the misprop is computed internally?

qubixes: I wanted to give it a little bit more leeway, since users might have their own computation of the prop_missing, so I don't want to put it at exactly 0.0 and 1.0. It's just to ensure that small rounding errors don't give errors.

PR #112, opened 2023-07-04
Raise an error if spec has unknown columns

### LINKED ISSUE #104: Unknown keys in MetaDataset specification are silently ignored
If the specification for creating a MetaDataset from a dataframe contains a key that does not correspond to any column in the dataframe, it is silently ignored, but should I think at least generate a warning to inform the user.
This:
```
df = pd.DataFrame([0,0,0], columns=['zeroes'])
var_spec = {  'id': {'unique': True} }
meta_dataset = MetaDataset.from_dataframe(df, spec=var_spec)
```
gives no warning or error.

qubixes:
Fixes #104

### DIFF
--- a/metasynth/dataset.py
+++ b/metasynth/dataset.py
@@ -117,6 +117,9 @@ class MetaDataset():
         else:
             spec = deepcopy(spec)
 
+        if set(list(spec)) - set(df.columns):
+            raise ValueError("Specifications found for column that were not found in the "
+                             f"dataset itself: {set(list(spec)) - set(df.columns)}")
         all_vars = []
         for col_name in df.columns:
             series = df[col_name]
--- a/metasynth/dataset.py
+++ b/metasynth/dataset.py
@@ -118,8 +118,9 @@ class MetaDataset():
             spec = deepcopy(spec)
 
         if set(list(spec)) - set(df.columns):
-            raise ValueError("Specifications found for column that were not found in the "
-                             f"dataset itself: {set(list(spec)) - set(df.columns)}")
+            raise ValueError("Argument 'spec' includes the specifications for column names that do "
+                             "not exist in the supplied dataframe:"
+                             f" '{set(list(spec)) - set(df.columns)}'")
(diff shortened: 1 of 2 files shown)


metasynth/dataset.py
```diff
@@ -117,6 +117,9 @@ def from_dataframe(cls,
         else:
             spec = deepcopy(spec)
 
+        if set(list(spec)) - set(df.columns):
```
vankesteren, line 120: This line looks great, set difference by just instantiating set objects

metasynth/dataset.py
```diff
@@ -117,6 +117,9 @@ def from_dataframe(cls,
...
+            raise ValueError("Specifications found for column that were not found in the "
+                             f"dataset itself: {set(list(spec)) - set(df.columns)}")
```
vankesteren, line 122: @Samuwhale do you have a suggestion for a better error message here? I am wondering if this is immediately understandable, but I'm finding it hard to decide on a better error message.

I checked what tidyverse does when loading a csv with a spec with a column that could not be found; it displays a warning (not an error) and it's the following message:
``` 
Warning message:                                                                     
The following named parsers don't match the column names: Pid 
```
I think that is kind of worse so maybe our message is ok 😃

tests/test_dataset.py
```diff
@@ -92,6 +92,10 @@ def check_dataset(dataset):
         print(name, dataset.descriptions[name])
         assert dataset.descriptions[name] == name
 
+    # Check whether non-columns raise an error
+    with pytest.raises(ValueError):
+        dataset = MetaDataset.from_dataframe(df, spec={"unicorn": {"prop_missing": 0.5}})
```
vankesteren, line 97: 🦄

vankesteren (approved):

metasynth/dataset.py
vankesteren: Feel free to merge and create a new issue out of this comment

Issue #120, opened 2023-07-26
Brainstorm about new name for "MetaDataset" object


vankesteren:
I'd like to brainstorm a bit about a new name for the `MetaDataset` object. Although it accurately represents some aspects of what this object means, in my mind, this name is creating some confusion. I think this might also become the case for new users. 

In my ideal world, the name should signal that this object:
- is like a "model" in scikit-learn language, with a `fit()` method (called `from_dataframe()`)
- contains information about each column
- is a (de)serializable representation of the model for the data, with (de)serialization handled by `to_json()` and `from_json()`.

I think `MetaDataset` does the third point well, but not the (important) first point and the (less important) second point. I'm not sure how important the third point is, I guess pretty important?

What could alternatives be?

- `GenerativeModel` or just `Model` (with preferred use as `metasynth.Model.from_dataframe(df)`)
- [sdv](https://docs.sdv.dev/sdv/single-table-data/modeling/synthesizers) calls this a `Synthesizer`, see also here:  https://github.com/sdv-dev/SDV/blob/master/sdv/single_table/base.py#L516

Do you have other ideas?

@qubixes @Samuwhale

Samuwhale:
Great point, it's actually something I wanted to bring up in the next meeting as I also felt the current terminology might be a bit confusing. Personally, I like the idea of referring to it as ```Blueprint```, as I feel like that is a universally understood term that fits here. I also like the term ```Model```,  but I'm worried that it might only make sense to those familiar with machine learning terminology (and confuse those that aren't). 

Curious to hear your thoughts.

qubixes:
What about `MetaFrame`? It's basically the metadata of the dataframe, and in some ways behaves like one (i.e. selecting columns).

vankesteren:
Agreed!

qubixes:
Done!

PR #123, opened 2023-07-31
Change MetaDataset -> MetaFrame

qubixes:
Also change from_dataframe -> fit_dataframe


vankesteren (approved):
Let's be consistent in what we assign MetaFrame to in the documentation. It is now sometimes called `dataset` sometimes `metadataset`, sometimes `mds`. Should we consistently call it `mf` or so? (just like it is common to call a DataFrame `df`)

vankesteren:
> Seems like there is an issue building polars on OS X/windows..

Not sure why it's trying to build polars from source on `macos-latest`?? Let's ignore for now.

vankesteren:
Fixes #120

PR #127, opened 2023-08-03
Create CITATION.cff

vankesteren:
It would be nice for metadata purposes (as we're a project which likes to do things with metadata) to include a CITATION.cff file.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,5 +1,5 @@
 cff-version: 1.2.0
-title: Metasynth
+title: MetaSynth
 message: >-
   If you use this software, please cite it using the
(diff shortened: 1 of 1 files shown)


qubixes (approved):
Nice!

qubixes (approved):

CITATION.cff
```diff
@@ -0,0 +1,42 @@
+cff-version: 1.2.0
+title: Metasynth
```
qubixes, line 2: Shouldn't we use `MetaSynth` to be consistent?

Samuwhale: For sure

PR #132, opened 2023-08-04
Regex improvement

### LINKED ISSUE #115: Regex distribution improvement
We had at several points several implementations of the RegexDistribution:

- A complicated, precise one that was developed to be really good at detecting regex but was very slow due to an inefficient search procedure
- The current implementation: a simpler faster greedy search method that we reverted to for user-friendliness in the tutorials/workshops we ran.

Ideally, we want both precise and fast. This issue is to start the conversation on how we would go about this. Maybe we can also search for other methods in other fields that are doing similar things and refer to those in the comments here?

### LINKED ISSUE #137: Distribution versioning
We need a way of versioning distributions if we update them

- version in decorator `@metadist(implements="core.uniform", version="2.0")`
- automatically use latest version
- add version in GMF file
- @Samuwhale made the nice suggestion that we can potentially move legacy distributions to a separate plugin
- think about how to keep contrib packages alive if they inherit from version 1. Maybe with `import as`?

qubixes:
I have been working on improving the regex distribution for a while now, and I think the result is pretty good!

I made a draft pull request, since I haven't finished docstrings, etc. Notes:

- It is based on polars regex matching, and it is *fast* (at least for structured strings). I used the titanic dataset, and it's <=0.2 seconds to run on my computer.
- I have changed the regex format:
  - `\d` -> `[0-9]`
  - `\w` -> REMOVED
  - `[a-zA-Z]` -> REMOVED
  - `.[]` -> REMOVED
  - `[abc]` -> Now has escaping. I realized that the old regex was broken for something like `[[]]`
 It is still possible to use regexes to initialize the distribution, but you need to escape now:
  - `r"[\\[][0-9]{3,3}[\\]]"` -> "[152]", "[623]", etc.
 
The regex distribution(s) now have a different parameter `score_thres`, now by default 0.1. This is a parameter that determines how many edge cases are tried to be fitted, lower values mean more precise, but more complex regexes. This removes a lot of junk that was created with the old regex fitting.

The length of the regex is not conserved anymore. This might be implemented in some way, but that might be somewhat complicated. Would be good to discuss if this is important, because it would be a breaking change.

The big difference is that now branching regexes are supported, which gives much better results. You can check an example `dist = Regex2Distribution.fit(pl.Series(["ja"]*100 + ["nee"]*100 + ["janee"]*100))`, to see the effect of this on (effectively) categorical variables.

Fixes #115, Fixes #137


vankesteren:
Nice, initial tests make this seem much better for unstructured text as well!

Two initial thoughts:

1. It's important that the exported json remains legible. Is it possible to have the regex_data in the json be something more legible than 
    ```js
     [
      (
       [
        {'left': [([], 1690)], 'right': [], 'regex': '[B]', 'center_count': 1690}, 
        {'left': [], 'right': [([], 1690)], 'regex': '[0-9]{5,5}', 'center_count': 0}
       ], 
       1690
      )
    ]
    ```
    for a column which is actually just 
    ```re
    B[0-9]{5}
    ```
    My first thought was that we could "pivot" this nested list somehow
    ```js
    {
       're': '[B][0-9]{5,5}', // legible regex
       'elements': ['[B]', '[0-9]{5,5}'], 
       'prop' : [1.0, 1.0], // instead of count
       'left' : ['', ''], // not sure what left and right do, do we need them?
       'right': ['', ''],
       'lprop': [1.0, 1.0],
       'rprop': [1.0, 1.0]
    }
    ```
       
2. Let's just replace the original regex distribution while we're at it. This regex2 thing can go IMO!

qubixes:
@vankesteren 

1. Yes, it can be improved. With the class structure I made, this was easy to implement, but it can be flatter. (half as flat)

- Counts -> probabilities, yes can be done, but needs some thought on normalization
- `left` and `right` can be nested regexes, so these will tend to sprawl with complicated regexes/unstructured text. They would not be strings, but are lists of strings.
- The first level should be a list, since it is a list of possible regexes (this is why categorical variables can be correct). It *can* be put  in a single tree technically, but that would necessitate an empty node somewhere. I'm not  sure if that is more intuitive.
- I'm not a fan of putting two values that are the same in there. Both are okay with me, but elements is probably okay to read for most people?

2. I'm okay with renaming `regex2 -> regex` and `regex -> old_regex`, but I think we should version the regex as `v2`. I think it is good practice to deprecate old functionality and not just change everything  so that old files cannot be worked with anymore. I think we should have a deprecation scheme, where old versions will be available for x minor/major versions. Defaults will take the latest version, but the old version should still work. Even though we're still in "alpha" stage, it will be good to *have* a deprecation scheme, just so that we don't just break everything downstream every version, and users will have reasonable warning times.

Samuwhale:
Hey! I did some tests myself too and like it so far, definitely think the generated results are an improvement over the old regex distribution.

Regarding the exported JSON format, I agree with @vankesteren's suggestion to make it more legible. We can discuss this further when we meet later today, perhaps.

With regards to the deprecation, I agree with you that we should be cautious with breaking functionality across versions, even if we are only in alpha. So I agree on discussing on a deprecation scheme (maybe in the next team meeting?).

Samuwhale (commented):
Hey! I did some tests myself too and like it so far, definitely think the generated results are an improvement over the old regex distribution.

Regarding the exported JSON format, I agree with @vankesteren's suggestion to make it more legible. We can discuss this further when we meet later today, perhaps.

With regards to the deprecation, I agree with you that we should be cautious with breaking functionality across versions, even if we are only in alpha. So I agree on discussing on a deprecation scheme (maybe in the next team meeting?).

vankesteren:
Oh yes, a really important point raised by #115: let's remove the dependency on scikit-learn. That would save a lot of disk space!

qubixes:
I have made a lot of modifications, and the following is now the output:

```{'regex_data': [{'regex': '[A-Z]{1,1}[0-9]{1,3}',
   'weights': [204, 200],
   'side_branches_before': [{'i_branch': 0, 'data': {'weight': 204}}],
   'side_branches_after': [{'i_branch': 0, 'data': {'weight': 4}},
    {'i_branch': 1, 'data': {'weight': 200}}]}]}```

I think this will be more readable than before? I'm keeping in the weights for now, since I think relative probabilities might be confusing as well? I could divide everything by the total number of records..

**Comment by qubixes** [comment]
@vankesteren @Samuwhale Let's have discussion about this branch after the weekend!

File: metasynth/distribution/__init__.py [hunk 1]
```diff
@@ -20,12 +20,14 @@
...
+from metasynth.distribution.regex import RegexDistribution, UniqueRegexDistribution
 from metasynth.distribution.na import NADistribution
 __all__ = [
     "MultinoulliDistribution", "UniformDistribution", "NormalDistribution",
     "LogNormalDistribution", "TruncatedNormalDistribution", "ExponentialDistribution",
     "DiscreteUniformDistribution", "PoissonDistribution", "UniqueKeyDistribution",
     "UniformDateDistribution", "UniformDateTimeDistribution", "UniformTimeDistribution",
-    "FakerDistribution", "RegexDistribution", "UniqueRegexDistribution",
+    "FakerDistribution", "RegexDistribution", "UniqueRegexDistribution", "RegexDistribution",
+    "UniqueRegexDistribution",
```
vankesteren, line 31: ? double stuff in here

metasynth/distribution/legacy/__init__.py
```diff
@@ -0,0 +1,5 @@
+from metasynth.distribution.legacy.regex import RegexDistribution as Regex1_0
+from metasynth.distribution.legacy.regex import UniqueRegexDistribution as UniqueRegex1_0
+
```
vankesteren, line 3: Maybe keep the Distribution part in the name?

metasynth/distribution/regex.py
vankesteren, line 1: I would need some in-person explanation of this file to be able to properly review it.

metasynth/provider.py
```diff
@@ -34,8 +34,10 @@
                                              PoissonDistribution,
                                              UniqueKeyDistribution)
 from metasynth.distribution.faker import FakerDistribution
-from metasynth.distribution.regex.base import (RegexDistribution,
-                                               UniqueRegexDistribution)
+from metasynth.distribution.regex import (RegexDistribution,
+                                          UniqueRegexDistribution)
+from metasynth.distribution.legacy import Regex1_0
+from metasynth.distribution.legacy import UniqueRegex1_0
```
vankesteren, line 40: If earlier name change is implemented, then also change it here

metasynth/provider.py
```diff
@@ -307,12 +352,17 @@ def from_dict(self, var_dict: dict[str, Any]) -> BaseDistribution:
...
+        version = var_dict["distribution"].get("version", "1.0")
+        var_type = var_dict["type"]
+        dist_class = self.find_distribution(dist_name, version=version, var_type=var_type)
+        return dist_class.from_dict(var_dict["distribution"])
+        # for dist_class in self._get_dist_list(var_type=var_dict["type"]) + [NADistribution]:
+            # if dist_class.implements == dist_name and dist_class.version == major_version:
+                # return dist_class.from_dict(var_dict["distribution"])
+        # raise ValueError(f"Cannot find distribution with name "
+                        #  f"'{var_dict['distribution']['implements']}'"
+                        #  f"and type '{var_dict['type']}'.")
```
vankesteren, line 365: remove :)

tests/test_regex.py
vankesteren, line 1: 🫢

tests/test_var.py
```diff
@@ -62,6 +61,9 @@ def check_similar(series_a, series_b):
         var_dict["distribution"].update({"implements": "unknown"})
         MetaVar.from_dict(var_dict)
 
+    print(var.to_dict())
+    print()
+    print(new_var.to_dict())
```
vankesteren, line 66: these are some debug statements yes?

vankesteren:
This is already more legible but still so nested!

```json
{'distribution': {
    'class_name': 'RegexDistribution',
    'implements': 'core.regex',
    'parameters': {
        'regex_data': [
            {
                'regex': '[b][a][n][a][n][a]',
                'side_branches_after': [{'data': {'regex': '[p][p][l][e]','side_branches_after': [{'data': {'weight': 2}, 'i_branch': 3}], 'side_branches_before': [], 'weights': [2, 2, 2, 2]}, 'i_branch': 1}, {'data': {'weight': 3}, 'i_branch': 5}], 'side_branches_before': [{'data': {'weight': 3}, 'i_branch': 0}, {'data': {'weight': 2}, 'i_branch': 1}], 'weights': [5, 3, 3, 3, 3, 3]}]},
                'provenance': 'builtin',
                'version': '2.0'
            },
 'dtype': 'Utf8',
 'name': 'fruits',
 'prop_missing': 0.0,
 'type': 'string'}
```

metasynth/distribution/legacy/__init__.py
qubixes: I abbreviated it because it was reaching the character limit for this line. I can reinstate it..

qubixes:
The code should be basically in a mergeable state. Obviously most of the logic is in the regexmodel package now.

vankesteren:
Don't have time for a full review yet! First major comment:

This does not work at all for a small example dataset, probably because there are only a few cases. Probably has to do with `count_thres=10` in the regexmodel package. Maybe we should scale this based on computational load (and only base this on privacy in plugins). 

Additionally, there should be a good fallback (e.g., "[a-zA-Z0-9]{4,4}" in the case below) if no nodes meet the `count_thres`, so that the package does not crash (see error at the bottom of this message) 


Reprex:
```python
import polars as pl
from metasynth import MetaFrame
from metasynth.distribution import RegexDistribution

# example dataframe from polars website
df = pl.DataFrame(
    {
        "ID": [1, 2, 3, 4, 5],
        "fruits": ["banana", "banana", "apple", "apple", "banana"],
        "B": [5, 4, 3, 2, 1],
        "cars": ["beetle", "audi", "beetle", "beetle", "beetle"],
        "optional": [28, 300, None, 2, -30],
        "roomnumber": ["a102", "b109", "a201", "b101", "c122"]
    }
)

# convert appropriate columns to categorical
df = df.with_columns([
    pl.col("fruits").cast(pl.Categorical),
    pl.col("cars").cast(pl.Categorical),
])

# set A to unique and B to not unique
spec_dict = {
    "ID": {"unique": True},
    "B": {"unique": False},
    "roomnumber": {"distribution": RegexDistribution}
}

# create MetaFrame
mf = MetaFrame.fit_dataframe(df, spec=spec_dict)

mf.synthesize(10)
```
Error:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Users\erikj\surfdrive\SoDa\projects\synthetic_data\metasyn\metasynth\dataset.py", line 277, in synthesize
    synth_dict = {var.name: var.draw_series(n) for var in self.meta_vars}
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\erikj\surfdrive\SoDa\projects\synthetic_data\metasyn\metasynth\dataset.py", line 277, in <dictcomp>
    synth_dict = {var.name: var.draw_series(n) for var in self.meta_vars}
                            ^^^^^^^^^^^^^^^^^^
  File "C:\Users\erikj\surfdrive\SoDa\projects\synthetic_data\metasyn\metasynth\var.py", line 244, in draw_series
    value_list = [self.draw() for _ in range(n)]
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\erikj\surfdrive\SoDa\projects\synthetic_data\metasyn\metasynth\var.py", line 244, in <listcomp>
    value_list = [self.draw() for _ in range(n)]
                  ^^^^^^^^^^^
  File "C:\Users\erikj\surfdrive\SoDa\projects\synthetic_data\metasyn\metasynth\var.py", line 226, in draw
    return self.distribution.draw()
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\erikj\surfdrive\SoDa\projects\synthetic_data\metasyn\metasynth\distribution\regex.py", line 57, in draw
    return self.regex_model.draw()
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\erikj\AppData\Roaming\Python\Python311\site-packages\regexmodel\model.py", line 211, in draw
    return self.regex_edge.draw()
           ^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\erikj\AppData\Roaming\Python\Python311\site-packages\regexmodel\datastructure.py", line 209, in draw
    return self.destination.draw()
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\erikj\AppData\Roaming\Python\Python311\site-packages\regexmodel\datastructure.py", line 142, in draw
    edge = np.random.choice(self.edges, p=self._probabilities)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "mtrand.pyx", line 934, in numpy.random.mtrand.RandomState.choice
ValueError: 'a' cannot be empty unless no samples are taken
```

qubixes:
@vankesteren I have made it so that regexmodel with throw an error in case it could not be fitted. It should now give you the default distribution.

qubixes:
@vankesteren I have made the default value of count_thres = None, and it will scale by default as #values/100. I have tested your example, and it should work.

metasynth/distribution/base.py
```diff
@@ -68,17 +69,20 @@ def draw(self) -> object:
     def draw_reset(self) -> None:
         """Reset the drawing of elements to start again."""
 
-    def __str__(self) -> str:
-        """Return an easy to read formatted string for the distribution."""
-        params_formatted = "\n".join(
+    @property
+    def _params_formatted(self):
```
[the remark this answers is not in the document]
vankesteren: Pay attention the dash may not be the right one because I'm on mobile, better just type this out yourself

metasynth/distribution/legacy/__init__.py
vankesteren: I think it's better to reinstate and then indent the import line differently. We want the ui for using this distribution to be as similar as possible to the original

metasynth/distribution/regex.py
```diff
@@ -0,0 +1,94 @@
...
+    of each of the branches in the regex is also recorded so that synthetic values can be
+    drawn from a better distribution.
+
+    Parameters
+    ----------
+    regex_data:
+        A valid input for the regex model. The two main ones is a serialized version
+        of the model, and the other is a regex that satisfies falls within the language
+        of the regexmodel package (which is a small subset of the python re package).
+    """
```
vankesteren, line 28: This should be improved, make it clear there are two options and potentially refer advanced users to the regexmodel package

metasynth/distribution/regex.py
```diff
@@ -0,0 +1,94 @@
...
+@metadist(implements="core.regex", var_type="string", version="2.0")
+class RegexDistribution(BaseDistribution):
+    """Distribution for structured strings.
+
+    Examples that this distribution should work reasonably for are: email,
+    ID's, telephone numbers, ip addresses, etc.
+
+    Valid regexes are ([a][b]|[c][d]), or [A-Z][0-9]{2,3}. During the fit, the probabilities
+    of each of the branches in the regex is also recorded so that synthetic values can be
+    drawn from a better distribution.
```
vankesteren, line 20: Suggestion: make this clearer and separate it from the data generation comment. For example:

Allowed regex elements are
- parentheses ()
- square brackets []
- single characters (e.g., a, \n, 3)
- the ranges [A-Z], [a-z], [0-9] (but not [B-F] or [4-7], then use [4567])
- repetition specification with two numbers {3,6} but not one {6}

For more information on the allowed regex subset, see the regexmodel package.

metasynth/distribution/regex.py
```diff
@@ -0,0 +1,94 @@
...
+
+from regexmodel import RegexModel, NotFittedError
+
+from metasynth.distribution.base import metadist, BaseDistribution, UniqueDistributionMixin
+
+
+@metadist(implements="core.regex", var_type="string", version="2.0")
+class RegexDistribution(BaseDistribution):
+    """Distribution for structured strings.
+
```
vankesteren, line 14: Start with a short description on what this is, and then go to examples

metasynth/distribution/regex.py
```diff
@@ -0,0 +1,94 @@
...
+    Examples that this distribution should work reasonably for are: email,
+    ID's, telephone numbers, ip addresses, etc.
+
+    Valid regexes are ([a][b]|[c][d]), or [A-Z][0-9]{2,3}. During the fit, the probabilities
+    of each of the branches in the regex is also recorded so that synthetic values can be
+    drawn from a better distribution.
+
```
vankesteren, line 21: We should also add here examples of how this distribution can be used 
- without fitting
- with the count_thres parameter

metasynth/distribution/regex.py
```diff
@@ -0,0 +1,94 @@
+"""Distribution for structured strings, using regexes."""
```
vankesteren, line 1: This file looks so clean now! Great

metasynth/provider.py
```diff
@@ -106,6 +115,9 @@ class BuiltinDistributionProvider(BaseDistributionProvider):
         UniformTimeDistribution,
         UniformDateTimeDistribution,
     ]
+    legacy_distributions = [
+        Regex1_0, UniqueRegex1_0
```
vankesteren, line 119: NB: Name change also here

metasynth/provider.py
```diff
@@ -226,8 +238,10 @@ def _find_best_fit(self, series: pl.Series, var_type: str,
                              f"that have unique == {unique}.")
         return dist_instances[np.argmin(dist_aic)]
 
-    def find_distribution(self, dist_name: str, privacy: BasePrivacy = BasicPrivacy(),
-                          ) -> type[BaseDistribution]:
+    def find_distribution(self,  # pylint: disable=too-many-branches
+                          dist_name: str, privacy: BasePrivacy = BasicPrivacy(),
+                          var_type: Optional[str] = None,
+                          version: Optional[str] = None) -> type[BaseDistribution]:
```
vankesteren, line 244:
```suggestion
    def find_distribution(self,  # pylint: disable=too-many-branches
                          dist_name: str, 
                          privacy: BasePrivacy = BasicPrivacy(),
                          var_type: Optional[str] = None,
                          version: Optional[str] = None) -> type[BaseDistribution]:
```

vankesteren: Minor formatting thingy. NB: I'm still on mobile so may look weird

vankesteren (changes requested):
Looks almost ready to merge!

I mainly have comments around the docstring. 

I'd also like to test it once more before merging, when I'm not on mobile. Maybe I can do that this afternoon

vankesteren:
just quickly tested on my computer! works 👍

qubixes:
@vankesteren I have improved the documentation, let me know if there is more to be adjusted!

vankesteren (approved):

pyproject.toml
```diff
@@ -79,6 +80,8 @@ ignore_missing_imports = true
 
 [tool.pylint.'MASTER']
 ignore-patterns="_version.py"
+ignore-paths = "^metasynth/distribution/legacy.*$"
+# ignore="metasynth.distribution.legacy"
```
qubixes, line 84:
```suggestion
```

Issue #130, opened 2023-08-04
update tutorial for new __str__ / __repr__ methods and new naming scheme 


Samuwhale:
Currently, the tutorial mentions that printing the metaframe results in a printed dictionary, but that's no longer true, it currently prints a readable string. The __repr__ _does_ however print a dictionary.

The tutorial should be updated accordingly.

![image](https://github.com/sodascience/metasynth/assets/22395533/18f2bc81-be65-4356-aa93-94e6c51b7f73)

Additionally, while the function names are adapted to the new naming scheme, the variable names aren't:

```python
new_meta_dataset = MetaFrame.from_json(file_path)
new_meta_dataset.synthesize(5)
```

These should be updated too.

PR #135, opened 2023-08-07
Tutorial update

Samuwhale:
Updates both tutorials, improved some passages in general to make them more understandable or more in line with the documentation. Also applied new naming scheme.

Tutorial should now be consistent with latest version of the package, as well as the conventions used for naming in the docs and readme. 

#130


qubixes (approved):
I have looked through it quickly, and it seems like a straightforward improvement!

Issue #121, opened 2023-07-31
Update the 'metadataset' GMF/JSON examples 


Samuwhale:
In the readme and on the docs, there are examples of GMF/JSON being given for the metadatasets. I believe these are outdated and should be updated.

PR #136, opened 2023-08-07
updated GMF examples to fit output of release 0.4

Samuwhale:
solves #121


qubixes (approved):
Good catch and approved!

PR #142, opened 2023-08-09
First version of CLI

vankesteren:
Added a command line interface, with the following help file:

```
usage: metasynth [-h] [-n NUM_ROWS] [-p] [-v] input output

Synthesize data from Generative Metadata Format .json file.

positional arguments:
  input                 Input file.
  output                Output file. File extension should be one of: .csv, .feather, .parquet, .pkl, .xlsx

options:
  -h, --help            show this help message and exit
  -n NUM_ROWS, --num_rows NUM_ROWS
                        Number of rows to synthesize
  -p, --print_only      print one-row data frame to console and exit
  -v, --version         show program's version number and exit
```

This will pave the way for stable, versioned docker containers which directly use this CLI, e.g., 
```docker
docker sodascience/metasynth:v0.4 metatdatafile.json outputfile.csv
```
Those containers could be automatically built whenever we release on GitHub as part of our CI

### DIFF
--- a/metasynth/__main__.py
+++ b/metasynth/__main__.py
@@ -2,6 +2,7 @@
 import argparse
 import pathlib
 import pickle
+import sys
 
 from metasynth._version import __version__
 from metasynth import MetaFrame
@@ -12,6 +13,7 @@ def main():
     parser = argparse.ArgumentParser(
         prog="metasynth",
         description="Synthesize data from Generative Metadata Format .json file.",
+        usage="%(prog)s synthesize [options] input output"
     )
     parser.add_argument(
         "input",
@@ -43,6 +45,12 @@ def main():
         version=f"%(prog)s {__version__}",
     )
 
+    # pop the shim subcommand and check that it is "synthesize"
+    subcommand = sys.argv.pop(1)
+    if subcommand != "synthesize":
+        parser.error(f"Invalid subcommand ({subcommand}).")
+
+    # parse the args without the shim subcommand
     args, _ = parser.parse_known_args()
 
     if not args.preview and not args.output:
--- a/metasynth/__main__.py
+++ b/metasynth/__main__.py
@@ -7,6 +7,7 @@ import sys
 from metasynth._version import __version__
 from metasynth import MetaFrame
 
+
 def main():
     """Parse arguments and generate synthetic data."""
     parser = argparse.ArgumentParser(
@@ -24,7 +25,7 @@ def main():
         type=pathlib.Path
     )
     parser.add_argument(
-        "-n", 
+        "-n",
         "--num_rows",
         help="Number of rows to synthesize",
         default=None,
--- a/metasynth/__main__.py
+++ b/metasynth/__main__.py
@@ -49,39 +49,38 @@ def main():
 
     print(args)
     # Create the metaframe from the json file
-    mf = MetaFrame.from_json(args.input)
+    meta_frame = MetaFrame.from_json(args.input)
 
     if args.print_only:
         # only print one row and exit
-        print(mf.synthesize(1))
+        print(meta_frame.synthesize(1))
         sys.exit(0)
 
     # Generate a data frame
     if args.num_rows is not None:
-        df = mf.synthesize(args.num_rows)
+        data_frame = meta_frame.synthesize(args.num_rows)
     else:
-        df = mf.synthesize(mf.n_rows)
+        data_frame = meta_frame.synthesize(meta_frame.n_rows)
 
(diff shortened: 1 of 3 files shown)


vankesteren:
Ok the tests now finally pass!!

metasynth/__main__.py
```diff
@@ -0,0 +1,88 @@
...
+    parser.add_argument(
+        "-n", "--num_rows",
+        help="number of rows to synthesize",
+        default=None,
+        type=int,
+        required=False
+    )
+    parser.add_argument(
+        "-p", "--print_only",
+        help="print six-row data frame to console and exit",
```
qubixes, line 37:
```suggestion
        help="preview six-row synthesized data frame to console and exit",
```

Maybe changing it to `--preview` makes sense?

tests/test_cli.py
qubixes, line 1: Nice test!

metasynth/__main__.py
```diff
@@ -0,0 +1,88 @@
...
+    elif args.output.suffix == ".feather":
+        data_frame.write_ipc(args.output)
+    elif args.output.suffix == ".parquet":
+        data_frame.write_parquet(args.output)
+    elif args.output.suffix == ".xlsx":
+        data_frame.write_excel(args.output)
+    elif args.output.suffix == ".pkl":
+        with open(args.output, "wb") as pkl_file:
+            pickle.dump(data_frame, file=pkl_file)
+            pkl_file.close()
```
qubixes, line 78:
```suggestion
```
The nice thing about context managers is that they close automatically!

metasynth/__main__.py
```diff
@@ -0,0 +1,88 @@
...
+    else:
+        parser.error(
+            f"Output file format ({args.output.suffix}) incorrect." +
```
qubixes, line 81:
```suggestion
            f"Unsupported output file format ({args.output.suffix})." +
```

metasynth/__main__.py
```diff
@@ -0,0 +1,88 @@
...
+            "Use .csv, .feather, .parquet, .pkl, or .xlsx."
+        )
+    sys.exit(0)
```
qubixes, line 84: Is there a reason to have this?

metasynth/__main__.py
```diff
@@ -0,0 +1,88 @@
...
+    if not args.print_only and not args.output:
+        parser.error("Output file is required.")
+
+    # Create the metaframe from the json file
+    meta_frame = MetaFrame.from_json(args.input)
+
+    if args.print_only:
+        # only print six rows and exit
+        print(meta_frame.synthesize(6))
+        sys.exit(0)
```
qubixes, line 58: Probably nicer to just return from the function?

metasynth/__main__.py
qubixes, line 1: I think it's a great idea to have a CLI interface. I do think we should have sub commands though. It is likely we would want to be able to create GMF files as well from the command line. `metasynth synthesize` and `metasynth create-metadata` perhaps?

qubixes (changes requested):
That's a nice feature indeed! I just have some small comments, looks very neat!

metasynth/__main__.py
vankesteren: Good point, I have two thoughts on it:

1. Creating metadata is an interactive thing. You need to load the data with the correct types, add a `spec`, and check whether everything runs correctly. Are we sure we want to allow people to do this with a simple CLI?
2. This will be some work, let's leave that for a next PR.

tests/test_cli.py
vankesteren: yeah, took some time to figure out but it works now!

metasynth/__main__.py
qubixes: > Good point, I have two thoughts on it:
> 
>     1. Creating metadata is an interactive thing. You need to load the data with the correct types, add a `spec`, and check whether everything runs correctly. Are we sure we want to allow people to do this with a simple CLI?
> 
>     2. This will be some work, let's leave that for a next PR.

I could see the `spec` being loaded in with a JSON file or something (That might be useful anyway with bigger datasets). Another sub command could be creating a visualization for example? I think in most cases we do want people to use it interactively, but regarding 2.): I'm not suggesting to make the `create-metadata` sub command  part of this PR, but if we want to support it in the future, it would be good to prepare for it with having sub commands (even if it's only one for now). Then we won't have to break the API in the future!

vankesteren: I've now added a "shim" subcommand, to keep the code simple.

usage is now 
```sh
metasynth synthesize input.json output.csv
```

PR #146, opened 2023-08-14
Add new unique faker distribution

qubixes:
I added a new mixin class for creating new unique distributions more easily.

### DIFF
--- a/metasynth/distribution/base.py
+++ b/metasynth/distribution/base.py
@@ -254,3 +254,27 @@ class ScipyDistribution(BaseDistribution):
 
     def _information_criterion(self, values):
         return 2 * self.n_par - 2 * np.sum(self.dist.logpdf(values))
+
+
+@metadist(is_unique=True)
+class UniqueMixin(BaseDistribution):
+    """Mixin class to make unique version of base distribution."""
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.key_set: set = set()
+
+    def draw_reset(self):
+        self.key_set = set()
+
+    def draw(self) -> object:
+        n_retry = 0
+        while n_retry < 1e5:
+            new_val = super().draw()
+            if new_val not in self.key_set:
+                self.key_set.add(new_val)
+                return new_val
+            n_retry += 1
+        raise ValueError(f"Failed to draw unique string after {n_retry} tries.")
+
+    def information_criterion(self, values):
+        return 9999999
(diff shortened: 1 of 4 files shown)


metasynth/distribution/base.py
```diff
@@ -254,3 +254,28 @@ def information_criterion(self, values):
...
+        while n_retry < 1e5:
+            new_val = super().draw()
+            if new_val not in self.key_set:
+                self.key_set.add(new_val)
+                return new_val
+            n_retry += 1
+        raise ValueError(f"Failed to draw unique string after {n_retry} tries.")
+
+    def information_criterion(self, values):
+        return 9999999
```
vankesteren, line 281: This is very elegant, we can use this for the other unique distributions too no?

metasynth/distribution/faker.py
```diff
@@ -59,3 +59,8 @@ def _param_schema(cls):
             "faker_type": {"type": "string"},
             "locale": {"type": "string"},
         }
+
+
+@metadist(implements="core.unique_faker", var_type="string")
```
vankesteren, line 64: Would it also work if we just do 
`@metadist(implements="core.unique_faker", var_type="string", is_unique="true")`
without subclassing `UniqueMixin`?

Would it make sense to call `UniqueMixin` something more descriptive about what method it overloads? Such as `UniqueSampler`, `UniqueDraws`, or even `UniqueRejectionSampler`?

vankesteren (approved):

metasynth/distribution/base.py
qubixes: Yes, that is indeed the point!

metasynth/distribution/faker.py
qubixes: The term "Mixin" is the kind of class this is, so I think that will help people familiar with the concept (or can find it on google). We can also call it `UniqueDistributionMixin` or `UniqueSamplerMixin`, etc. That might be the most descriptive. It's basically a copy from the `UniqueRegexDistribution`, but I don't want to mess with that right now, while I'm messing with the regex distributions...

It is probably possible to implement a decorator like that:

https://stackoverflow.com/questions/38405619/how-to-mixin-behavior-using-class-decorators-in-python

But honestly, I think that will just drive up the complexity a little too much..

vankesteren: Agreed with everything. Either of those names you suggested are good!

qubixes:
@vankesteren Should we already merge or should we work on documenting the feature?

vankesteren:
> work on documenting the feature?

What do you mean? As in: developer documentation?

qubixes:
Yes, I suppose mostly developer documentation.

qubixes:
But maybe we should organize that in one go instead.

vankesteren:
Yeah, let's merge!

Issue #125, opened 2023-08-03
Updates for print methods


vankesteren:
With the new print method for the `MetaFrame` object there are still two things that should be fixed:

- [ ] Not all distributions print the same
- [ ] Pretty printing is only done when you explicitly call `print()` on the object, not when you just type the object name in the REPL.

## Not all distributions print the same

Some distributions still print as a dict (e.g., `multinoulli`), and some print nicely as a nested bulletpoint list (e.g., `unique_key`):

_Multinoulli_
```
{'implements': 'core.multinoulli', 'provenance': 'builtin', 'class_name': 'MultinoulliDistribution', 'parameters': {'labels': array(['apple', 'banana'], dtype='<U6'), 'probs': array([0.4, 0.6])}}`)
```
_Unique key_
```
- Distribution:
        - Type: core.unique_key
        - Provenance: builtin
        - Parameters:
                - low: 1
                - consecutive: 1
```

A reproducible example of this: 

```py
import polars as pl
from metasynth import MetaFrame

# example dataframe from polars website
df = pl.DataFrame(
    {
        "ID": [1, 2, 3, 4, 5],
        "fruits": ["banana", "banana", "apple", "apple", "banana"],
        "B": [5, 4, 3, 2, 1],
        "cars": ["beetle", "audi", "beetle", "beetle", "beetle"],
        "optional": [28, 300, None, 2, -30],
    }
)

# convert appropriate columns to categorical
df = df.with_columns([
    pl.col("fruits").cast(pl.Categorical),
    pl.col("cars").cast(pl.Categorical),
])

# set A to unique and B to not unique
spec_dict = {
    "ID": {"unique": True},
    "B": {"unique": False}
}

# create metadataset
mf = MetaFrame.fit_dataframe(df, spec=spec_dict)

```

```
# Rows: 5
# Columns: 5

Column 1: "ID"
- Variable Type: discrete
- Data Type: Int64
- Proportion of Missing Values: 0.0000
- Distribution:
        - Type: core.unique_key
        - Provenance: builtin
        - Parameters:
                - low: 1
                - consecutive: 1


Column 2: "fruits"
- Variable Type: categorical
- Data Type: Categorical
- Proportion of Missing Values: 0.0000
- Distribution:
[...]

PR #147, opened 2023-08-15
fixed categorical and faker dists not being in line with new formatted print methods

Samuwhale:
Every distribution now SHOULD print the same, this solves #125.

Tests for faker have been updated accordingly.


qubixes (approved):
Thanks!

PR #152, opened 2023-08-17
Allow synthesis with no arguments

### LINKED ISSUE #126: `MetaFrame.n_rows` is `Optional | int`
This gives type annotation errors for example when you try to synthesize data with the same number of rows as the real data. Can we set the `n_rows` property to be of type `int` always?

Example:
```py
df_synth = mf.synthesize(mf.n_rows)
```

Then `mypy` gives the following error:
```
Argument 1 to "synthesize" of "MetaFrame" has incompatible type "Optional[int]"; expected "int"  [arg-type]
```

qubixes:
Resolves #126


vankesteren (approved):
👍👍👍

PR #153, opened 2023-08-17
Extend range of multinoulli distribution

### LINKED ISSUE #144: Add DiscreteMultinoulliDistribution
Multinoulli distribution should also be possible for discrete variables, happens a lot when dummy coding (0/1) or when value labels of a categorical variable are in another table

qubixes:
String and discrete are now supported as well. A surprisingly small change indeed.

Fixes #144

### DIFF
--- a/metasynth/distribution/categorical.py
+++ b/metasynth/distribution/categorical.py
@@ -64,28 +64,70 @@ class MultinoulliDistribution(BaseDistribution):
         labels, counts = np.unique(series, return_counts=True)
         log_lik = 0.0
         pdict = dict(zip(self.labels, self.probs))
+        # Check type of variable and act accordingly.
         if len(self.labels) > 1 and isinstance(self.labels[0], (int, np.int_)):
-            # Do add-one smoothing, assume count of values is equal to one in initial fit.
-            alpha = 1
-            num_event = len(series)
-            num_tot_cat = np.max(self.labels)-np.min(self.labels)+1
-            num_zero_cat = num_tot_cat - len(self.labels)
-            theta = (self.probs*num_event+alpha)/(num_event+num_tot_cat*alpha)
-            theta_zero = alpha/(num_event+num_tot_cat*alpha)
-            theta_sum = np.sum(theta) + num_zero_cat*theta_zero
-            theta_dict = dict(zip(self.labels, theta))
+            log_lik = self._log_like_int(series, labels, counts)
+            # See docstring of log_like_int method.
+            n_parameters = len(self.probs)-1 + len(self.probs)
+        else:
             for lab, count in zip(labels, counts):
-                cur_theta = theta_dict.get(lab, theta_zero)
-                cur_prob = cur_theta/theta_sum
-                log_lik += count * np.log(cur_prob)
-            return 2*(2*len(self.probs)-1) - 2*log_lik
+                # account for missing values / missing categories
+                # by setting default of .get to 1 (add log(1)=0 to log_lik)
+                log_lik += count * np.log(pdict.get(lab, 1))
+            n_parameters = len(self.probs)-1
 
+        return 2*n_parameters - 2*log_lik
+
+    def _log_like_int(
+            self,
+            series: pl.Series,
+            labels: npt.NDArray[np.int_],
+            counts: npt.NDArray[np.int_],
+            alpha: float = 1.0,
+            ) -> float:
+        """Information criterion for the integer variant of the multinoulli distribution.
+
+        This information criterion is a little bit different from the AIC, which is why
+        the method is split into its own method. The reason is that with the default IC,
+        the multinomial distribution is unfairly favored compared to the uniform distribution
+        in particular (but really any discrete distribution). Assume for instance that we have a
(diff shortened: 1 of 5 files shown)


qubixes:
@vankesteren Ready to merge?

metasynth/distribution/categorical.py
```diff
@@ -42,33 +42,52 @@ def __init__(self, labels: Union[npt.NDArray[np.str_], list[str]],
...
+            num_zero_cat = num_tot_cat - len(self.labels)
+            theta = (self.probs*num_event+alpha)/(num_event+num_tot_cat*alpha)
+            theta_zero = alpha/(num_event+num_tot_cat*alpha)
+            theta_sum = np.sum(theta) + num_zero_cat*theta_zero
+            theta_dict = dict(zip(self.labels, theta))
+            for lab, count in zip(labels, counts):
+                cur_theta = theta_dict.get(lab, theta_zero)
+                cur_prob = cur_theta/theta_sum
+                log_lik += count * np.log(cur_prob)
+            return 2*(2*len(self.probs)-1) - 2*log_lik
```
vankesteren, line 81: This should be actual AIC, i.e.:

```suggestion
            return 2*(len(self.probs)-1) - 2*log_lik
```

metasynth/distribution/categorical.py
```diff
@@ -42,33 +42,52 @@ def __init__(self, labels: Union[npt.NDArray[np.str_], list[str]],
...
                               values: Union[pd.Series, pl.Series, npt.NDArray[np.str_]]
                               ) -> float:
-        values_array = np.array(values, dtype=str)
-        labels, counts = np.unique(values_array, return_counts=True)
+        series = self._to_series(values)
+        labels, counts = np.unique(series, return_counts=True)
         log_lik = 0.0
         pdict = dict(zip(self.labels, self.probs))
+        if len(self.labels) > 1 and isinstance(self.labels[0], (int, np.int_)):
+            # Do add-one smoothing, assume count of values is equal to one in initial fit.
```
vankesteren, line 68: Why add-one smoothing and not add-half smoothing?

metasynth/distribution/categorical.py
```diff
@@ -42,33 +42,52 @@ def __init__(self, labels: Union[npt.NDArray[np.str_], list[str]],
...
+            return 2*(2*len(self.probs)-1) - 2*log_lik
+
         for lab, count in zip(labels, counts):
             # account for missing values / missing categories
             # by setting default of .get to 1 (add log(1)=0 to log_lik)
             log_lik += count * np.log(pdict.get(lab, 1))
-        return 2*(len(self.probs) - 1) - 2 * log_lik
+
+            # delta = max(1, np.max(self.labels)-np.min(self.labels))
+            # return 2*(len(self.probs)+len(series)*np.log(delta)) - 2*log_lik
```
vankesteren, line 89:
```suggestion

```

metasynth/distribution/categorical.py
```diff
@@ -42,33 +42,52 @@ def __init__(self, labels: Union[npt.NDArray[np.str_], list[str]],
...
+        return 2*(2*len(self.probs)-1) - 2 * log_lik
```
vankesteren, line 90: Here too!
```suggestion
        return 2*(len(self.probs)-1) - 2 * log_lik
```

metasynth/distribution/categorical.py
vankesteren: This comment should also be more explicit, why are we doing this, what effects does it have? or documented better in the docstring.

vankesteren (changes requested):

metasynth/distribution/categorical.py
qubixes: I kind of agree, but can't we see both the label and the probability as a parameter? If you compare to a uniform distribution [a, b], which has according to our definitions two parameters, then the multinoulli would have 2? parameters as well, except that in reality it has effectively [a, p_a, b, p_b] parameters, where you can have p_a == p_b, but don't have to adhere to this. Personally, I think this should be classified as 4 parameters, at least in the small count/label limit. In the limit of large n, this factor has limited influence as well. Another consideration that I think generally multinoulli a little bit more discouraged than perhaps normally, since we don't want too many parameters in our distributions.

metasynth/distribution/categorical.py
qubixes: > Why add-one smoothing and not add-half smoothing?

I don't really have any solid grounds on it, except it being argued in the Wikipedia page. Again, since I think the multinoulli distribution is more of a "last resort" type of distribution, if 1 is reasonable it might be preferable? If you think 1/2 is better defensible than 1, then we can put it to 1/2, but from what I read it seems like it depends on your assumption/taste.

vankesteren:
After discussion, the solution will be: 

Extensive docstring in information_criterion (and potentially log-likelihood if that becomes a separate method) with:
- That this is a PSEUDO-ic
- Why this formulation of the IC (choice vs. uniform distribution)
- What changes have been made relative to the "normal" IC of the multinoulli
  - smoothing
  - extra multiplier in IC formula
- How this can be edited to change behaviour

vankesteren:
Yes!

Issue #151, opened 2023-08-17
Version Generative Metadata Format based on metasynth python package


vankesteren:
__Introduction__
Since #103 we have not had static json schema files available, and we are validating the jsons internally, with dynamically generated files (see [here](https://github.com/sodascience/metasynth/blob/main/metasynth/validation.py)). We did this because it is a lot of effort to manually write these JSON schemas and to keep updating it every time we make changes to this python package.

However, I think it is still good to have a versioned JSON schema available somewhere, for FAIRness; mainly interoperability with other software such as existing json validators and [this thing](https://json-schema-faker.js.org/). 

__Proposal__
Since we already have a lot of the necessary infrastructure in place, my proposal is as follows: 

- Make this python package the official source for the generative metadata format
- Align the version of the json schema with the version of this package
- With every tagged release, we automatically generate and push the core json schema to the [GMF repository](https://github.com/sodascience/generative_metadata_format) using a github action. I've already done a little test that works [here](https://github.com/vankesteren/metasynth/actions/runs/5878801805); see the artifact that is produced.

This aligns well with our recent changes, adding a cli (#142) and the inclusion of a docker container (#150).

__Changes__
Changes are mainly in the validation script (using `__version__` for the schema base). We could include generating the schema in the CLI, which would make the github action simpler and allow us to (re)generate previous schema versions using our versioned docker containers. The most work will be in creating the github action to push to a different repo (which we already do for our website so that's definitely possible)

The GMF repository will be kind of "broken", as the versions will change. I think this is okay.

__Out of scope__
We should think about how to include / deal with plugins at a later date.

qubixes:
I think this is too far down the priority list to solve. Personally, I think the GMF version should be separate from the metasyn versions, even though arguably this is a little harder to keep track of. The GMF repo should be updated though, but then the issue should be there and not here.

PR #154, opened 2023-08-17
Add jsonschema entrypoint to CLI

vankesteren:
- Overhauled argument parsing for CLI
- Better error message for top-level CLI call
- Automatic dispatch to different entrypoints in `__main__.py` file

### DIFF
--- a/metasynth/__main__.py
+++ b/metasynth/__main__.py
@@ -1,17 +1,17 @@
 """CLI for generating synthetic data frames from a metasynth .json file."""
 import argparse
+import json
 import pathlib
 import pickle
 import sys
-import json
 
-try:
+try:  # Python < 3.10 (backport)
     from importlib_metadata import entry_points
 except ImportError:
-    from importlib.metadata import entry_points  # type: ignore
+    from importlib.metadata import entry_points  # type: ignore [assignment]
 
-from metasynth._version import version_tuple
 from metasynth import MetaFrame
+from metasynth._version import version_tuple
 from metasynth.validation import create_schema
 
 SEMVER = f"{version_tuple[0]}.{version_tuple[1]}.{version_tuple[2]}"
@@ -54,30 +54,30 @@ def synthesize() -> None:
     """Program to generate synthetic data."""
     parser = argparse.ArgumentParser(
         prog="metasynth synthesize",
-        description="Synthesize data from Generative Metadata Format .json file."
+        description="Synthesize data from Generative Metadata Format .json file.",
     )
     parser.add_argument(
         "input",
         help="input file; typically .json adhering to the Generative Metadata Format",
-        type=pathlib.Path
+        type=pathlib.Path,
     )
     parser.add_argument(
         "output",
         help="output file (.csv, .feather, .parquet, .pkl, or .xlsx)",
         nargs="?",
-        type=pathlib.Path
+        type=pathlib.Path,
     )
     parser.add_argument(
         "-n", "--num_rows",
         help="number of rows to synthesize",
         default=None,
         type=int,
-        required=False
+        required=False,
     )
     parser.add_argument(
         "-p", "--preview",
         help="preview six-row synthesized data frame in console and exit",
-        action="store_true"
+        action="store_true",
     )
 
     # parse the args without the subcommand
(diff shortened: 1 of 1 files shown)


vankesteren:
Part of #151

metasynth/__main__.py
```diff
@@ -1,56 +1,86 @@
...
+try:  # Python < 3.10 (backport)
+    from importlib_metadata import entry_points
+except ImportError:
+    from importlib.metadata import entry_points  # type: ignore [assignment]
+
 from metasynth import MetaFrame
+from metasynth._version import version_tuple
+from metasynth.validation import create_schema
+
+SEMVER = f"{version_tuple[0]}.{version_tuple[1]}.{version_tuple[2]}"
```
qubixes, line 17: Why not use the actual version?

metasynth/__main__.py
```diff
@@ -1,56 +1,86 @@
...
+    -h, --help    - display this help file and exit
+"""
+
+ENTRYPOINTS = ["synthesize", "jsonschema"]
+
 
+def main() -> None:
+    """CLI pointing to different entrypoints."""
+    # show help by default, else consume first argument
+    subcommand = "--help" if len(sys.argv) < 2 else sys.argv.pop(1)
```
qubixes, line 39: It looks like argparse has functionality for sub-commands natively. Or is there a reason to do it manually?

https://docs.python.org/3/library/argparse.html#sub-commands

metasynth/__main__.py
```diff
@@ -80,13 +110,55 @@ def main():
...
+        "plugins",
+        help="Plugins to include in the generated schema (default none)",
+        nargs="*",
+    )
+
+    parser.add_argument(
+        "-a", "--avail",
+        help="display available modules and quit",
+        action="store_true",
+    )
```
qubixes, line 139: I think this should be `-l/--list`, which seems to be the standard.

metasynth/__main__.py
```diff
@@ -80,13 +110,55 @@ def main():
...
+def jsonschema() -> None:
+    """Program to generate json schema from dist providers."""
+    parser = argparse.ArgumentParser(
+        prog="metasynth jsonschema",
+        description="Create Generative Metadata Format jsonschema and print to console.",
+    )
+
+    parser.add_argument(
+        "plugins",
+        help="Plugins to include in the generated schema (default none)",
```
qubixes, line 131: This should be default builtin, no?

metasynth/__main__.py
```diff
@@ -80,13 +110,55 @@ def main():
...
+    )
+
+    # parse the args without the subcommand
+    args, _ = parser.parse_known_args()
+
+    # deduplicated list of plugins for schema
+    plugins_avail = {entry.name for entry in entry_points(group="metasynth.distribution_provider")}
+
+    if args.avail:
+        _ = [print(a) for a in plugins_avail if a != "builtin"]
```
qubixes, line 148: I think it's clearer if it's just a for-loop.

metasynth/__main__.py
```diff
@@ -68,7 +98,7 @@ def main():
     if args.num_rows is not None:
         data_frame = meta_frame.synthesize(args.num_rows)
     else:
-        data_frame = meta_frame.synthesize(meta_frame.n_rows)
+        data_frame = meta_frame.synthesize(meta_frame.n_rows)  # type: ignore [arg-type]
```
qubixes, line 101: Why is here an ignored type check? With the latest version, should it not accept None values?

qubixes (changes requested):
That's a very nice addition! I'm still debating whether it should be `metasynth jsonschema`, or `metasynth schema`, maybe the second also covers it? I didn't know myself about the sub parsers, but I think it would fit well in our project to use them!

metasynth/__main__.py
vankesteren: won't fix in this PR if that's ok. Could not get the help file to display how I want. We can fix it if we want to add a new entrypoint or this file becomes more complex.

metasynth/__main__.py
```diff
@@ -80,13 +108,58 @@ def main():
...
+        action="store_true",
+    )
+
+    # parse the args without the subcommand
+    args, _ = parser.parse_known_args()
+
+    # deduplicated list of plugins for schema
+    plugins_avail = {entry.name for entry in entry_points(group="metasynth.distribution_provider")}
+
+    if args.list:
```
qubixes, line 146: I think for now it's okay to make "builtin" special, but I think if other packages are coming up, we might want to not make it special.

qubixes (approved):
Looks good!!

Issue #139, opened 2023-08-07
Documentation: Add 'advanced' arguments to mf generation in user guide on docs


Samuwhale:

vankesteren:
See also #140

Samuwhale:
Fixed in #155

PR #155, opened 2023-08-18
Update docs to have more advanced features in metadata generation page

Samuwhale:
Updates to docs in line with #139.

Changed:
- Added FAQ entry for unique value warning 
- Updated generating metadata page with advanced features
- Added some links to API reference
- Added a warning on the developer guide detailed overview to communicate that page might be out of date


vankesteren (approved):
nice work!

Issue #160, opened 2023-09-08
Add an additional function named 'export'


Samuwhale:
I discussed with @qubixes and we agreed on more consistent usage of the term 'GMF file' as opposed to the longer, and sometimes more vague 'MetaFrame exported to JSON', but in line with this, it might be nice to add a new `export` function.

It could simply be a wrapper function around `to_json`, it does not have to replace it. 

It's a very simple addition, but would avoid having to use confusing sentences in the docs along the lines of 'you can export a metaframe to a gmf file with `to_json`' and instead facilitate more a more natural 'you can export a metaframe to a gmf file with `export`'.

vankesteren:
great idea.

PR #162, opened 2023-09-11
added 'export' wrapper function to 'to_json'

Samuwhale:
solves #160

### DIFF
--- a/metasynth/dataset.py
+++ b/metasynth/dataset.py
@@ -212,7 +212,7 @@ class MetaFrame():
 
     def to_json(self, fp: Union[pathlib.Path, str],
                 validate: bool = True) -> None:
-        """Write the MetaSynth dataset to a JSON file.
+        """Serialize and export the MetaSynth dataset to a JSON file.
 
         Optional validation against a JSON schema included in the package.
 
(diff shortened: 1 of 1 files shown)


metasynth/dataset.py
```diff
@@ -212,7 +212,7 @@ def descriptions(
 
     def to_json(self, fp: Union[pathlib.Path, str],
                 validate: bool = True) -> None:
-        """Write the MetaSynth dataset to a JSON file.
+        """Serialize and export the MetaSynth dataset to a JSON file.
```
qubixes, line 215:
```suggestion
        """Serialize and export the MetaFrame to a JSON file.
```

metasynth/dataset.py
```diff
@@ -229,6 +229,21 @@ def to_json(self, fp: Union[pathlib.Path, str],
         with open(fp, "w", encoding="utf-8") as f:
             json.dump(self_dict, f, indent=4)
 
+    def export(self, fp: Union[pathlib.Path, str],
+               validate: bool = True) -> None:
+        """Serialize and export the MetaSynth dataset to a JSON file.
```
qubixes, line 234:
```suggestion
        """Serialize and export the MetaFrame to a JSON file.
```

metasynth/dataset.py
```diff
@@ -229,6 +229,21 @@ def to_json(self, fp: Union[pathlib.Path, str],
...
+
+        Optional validation against a JSON schema included in the package.
```
qubixes, line 236:
```suggestion
        Optionally, the exported JSON file is validated against the JSON schema(s) included in the package.
```

metasynth/dataset.py
```diff
@@ -229,6 +229,21 @@ def to_json(self, fp: Union[pathlib.Path, str],
...
+
+        Parameters
+        ----------
+        fp:
+            File to write the dataset to.
+        validate:
+            Validate the JSON file with a schema.
+        """
+        self.to_json(fp, validate)
```
qubixes, line 245: I think if we want to talk about the `export` in the documentation (instead of the `to_json` file), then we should also put the implementation here, and make the `to_json` method the wrapper function.

qubixes (changes requested):
Thanks! I have some small suggestions/comments, feel free to adjust/disagree.

Issue #163, opened 2023-09-11
update readme to reflect changes to docs with regards to the pipeline (estimation, de-/serialization and generation)


Samuwhale:
Now that the new docs (with more technical information as opposed to the prior more abstract one) have been merged, it's good to update the readme to reflect these changes.

PR #173, opened 2023-09-18
Updated readme to reflect current docs state

Samuwhale:
- Condensed irrelevant information
- referred to the docs in the introduction
- Adapted the pipeline as now explained on the docs
- Changed 'to_json' (function) to 'export'
- Mentioned GMF files more (as opposed to .JSON files)
- improved the warning message on 'unique' values

Improved warning message + grammatical error fixes

Solves #163

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,142 +8,30 @@
 # MetaSynth
 MetaSynth is a Python package designed to generate tabular synthetic data for rigorous code testing and reproducibility.
 
-The package has two main functionalities. First, it allows for the **creation of metadata** from an input dataset. This metadata describes the overarching structure and traits of the dataset. Second, MetaSynth allows for **generation of synthetic data** that aligns with this metadata. Instead of relying on the original dataset, the synthetic data is produced using the metadata. This approach ensures that the synthetic dataset remains separate and independent from any sensitive source data. Researchers and data owners can leverage this capability to generate and share synthetic versions of their sensitive data, mitigating privacy concerns. Furthermore, this separation between metadata and original data promotes reproducibility, as the metadata file can be easily shared and used to generate consistent synthetic datasets.
+The package has three main functionalities:
 
+1. **Estimation**: MetaSynth can create a MetaFrame, from a dataset. A MetaFrame is essentially a fitted model that characterizes the structure of the original dataset without storing actual values. It captures individual distributions and features, enabling generation of synthetic data based on these MetaFrames and can be seen as (statistical) metadata.
+2. **Serialization**: MetaSynth can export a MetaFrame into an easy to read JSON file, allowing users to audit, understand, and modify their data generation model.
+3. **Generation**: MetaSynth can generate synthetic data based on a MetaFrame. The synthetic data produced solely depends on the MetaFrame, thereby maintaining a critical separation between the original sensitive data and the synthetic data generated. The generated synthetic data, emulates the original data's format and plausibility at the individual record level and attempts to reproduce marginal (univariate) distributions where possible. Generated values are based on the observed distributions while adding a degree of variance and smoothing. The generated data does **not** aim to preserve the relationships between variables. The frequency of missing values and their codes are maintained in the synthetically-augmented dataset. 
 
-## Features
-### Generating metadata from a dataset
-MetaSynth can generate metadata from any given dataset (provided as polars or pandas dataframe) in the form of a MetaFrame. A MetaFrame encapsulates the structure and characteristics of each column in the original dataset (including their names, variable types, data types, proportion of missing values and distribution specifications) and serves as complete recipes for generating new synthetic data. 
+Researchers and data owners can use MetaSynth to generate and share synthetic versions of their sensitive datasets, mitigating privacy concerns. Additionally, MetaSynth facilitates transparency and reproducibility, by allowing the underlying MetaFrames to be exported and shared. Other researchers can use these to regenerate consistent synthetic datasets, validating published work without requiring sensitive data.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -8,142 +8,30 @@
...
 
-![Synthetic_data_generation](docs/source/images/flow_synthetic_data_generation.png)
+Researchers and data owners can use MetaSynth to generate and share synthetic versions of their sensitive datasets, mitigating privacy concerns. Additionally, MetaSynth facilitates transparency and reproducibility, by allowing the underlying MetaFrames to be exported and shared. Other researchers can use these to regenerate consistent synthetic datasets, validating published work without requiring sensitive data.
```
qubixes, line 17: I think I would rather have it at before the list, since it seems the most important part of the README.

README.md
```diff
@@ -8,142 +8,30 @@
...
+-   **MetaFrame Generation**: MetaSynth allows the creation of a MetaFrame from a dataset provided as a Polars or Pandas DataFrame.
+    MetaFrames includes key characteristics such as *variable names*, *data types*, *the percentage of missing values*, and *distribution parameters*.
+-   **Exporting MetaFrames**: MetaSynth can export and import MetaFrames to GMF files. These are JSON files that follow the easy to read and understand [Generative Metadata Format (GMF)](https://github.com/sodascience/generative_metadata_format).
+-   **Synthetic Data Generation**: MetaSynth allows for the generation of a polars DataFrame with synthetic data that resembles the original data.
+-   **Distribution Fitting**: MetaSynth allows for manual and automatic distribution fitting.
+-   **Data Type Support**: MetaSynth supports generating synthetic data for a variety of common data types including `categorical`, `string`, `integer`, `float`, `date`, `time`, and `datetime`.
+-   **Integration with Faker**: MetaSynth integrates with the [faker](https://github.com/joke2k/faker) package, a Python library for generating fake data such as names and emails. Allowing for more realistic synthetic data.
```
qubixes, line 28: I would add more private also on top of realistic, since it will contain no information of the original dataset.

README.md
```diff
@@ -8,142 +8,30 @@
...
+-   **Structured String Detection**: MetaSynth identifies structured strings within your dataset, which can include formatted text,
+    codes, identifiers, or any string that follows a specific pattern.
+-   **Handling Unique Values**: MetaSynth can identify and process variables with unique values or keys in the data, preserving their uniqueness in the synthetic dataset, which is crucial for generating synthetic data that maintains the characteristics of the original
+    dataset.
```
qubixes, line 32: I think the second part of the sentence doesn't add too much.
```suggestion
-   **Handling Unique Values**: MetaSynth can identify and process variables with unique values or keys in the data, preserving their uniqueness in the synthetic dataset.
```

README.md
```diff
@@ -213,28 +101,30 @@ be easily converted back to a Pandas DataFrame by using `df_pandas = df_polars.t
...
 > ```
 
-##### 3. We can export this MetaFrame to a .JSON file using:
+#### (De)serialization: Exporting and importing a MetaFrame 
+_Note that exporting and importing is optional. You can generate synthetic data from **any** loaded MetaFrame, whether that be through importing a GMF file or generating a MetaFrame from an original DataFrame._
+
+##### 3. We can export this MetaFrame to a GMF file using:
 
 ```python
 #export MetaFrame
```
qubixes, line 116:
```suggestion
# export MetaFrame
```

README.md
```diff
@@ -213,28 +101,30 @@ be easily converted back to a Pandas DataFrame by using `df_pandas = df_polars.t
 mf = MetaFrame.fit_dataframe(df)
 ```
 
-> Note: if at this point you get the following warning about a potential unique variable, do not worry, it is safe to continue.
+> Note: If you encounter a warning about `PassengerId` not being set as unique, you can safely ignore it and proceed. This warning occurs because `PassengerId` appears to contain unique values, but is not explicitly marked as a unique column. To remove the warning, you can set `PassengerId` to be a unique column. Our documentation explains how to do this when generating Metaframes: [Set Columns as Unique](https://metasynth.readthedocs.io/en/latest/usage/generating_metaframes.html#optional-parameters).
```
qubixes, line 104: In this case you will always get the warning, right? That should then be reflected in this sentence. I think it can be shorter like in the original too. The link is really nice though!

README.md
```diff
@@ -8,142 +8,30 @@
...
-
-## Features
-### Generating metadata from a dataset
-MetaSynth can generate metadata from any given dataset (provided as polars or pandas dataframe) in the form of a MetaFrame. A MetaFrame encapsulates the structure and characteristics of each column in the original dataset (including their names, variable types, data types, proportion of missing values and distribution specifications) and serves as complete recipes for generating new synthetic data. 
-
-MetaFrames follow the GMF standard, [Generative Metadata Format (GMF)](https://github.com/sodascience/generative_metadata_format) and as such are designed to be easy to read. MetaFrames can be exported as .JSON file allowing for manual and automatic editing, as well as easy sharing.
-
-![Metadata_generation_flowchart](docs/source/images/flow_metadata_generation.png)
-
-<details>
```
qubixes, line 22: What is the reason you removed this example? Maybe it might be nice to at least link to an example?

qubixes (changes requested):
Thanks! I didn't realize this was still in the pipeline! Let's try to get this merged as soon as possible, so that we can continue with the rename.

README.md
Samuwhale: Something like 'allowing for synthetic data that is formatted realistically, while retaining privacy'?

README.md
Samuwhale: added it back in now, in a nicer place (in the key features overview)

Samuwhale:
Should be good to go!

qubixes (approved):
Approved!

Issue #179, opened 2023-09-27
Add rename to FAQ


qubixes:
It might be nice to have a section on the rename of metasynth to the FAQ (of metasyn!).

PR #181, opened 2023-09-29
Added section on the name change in the FAQ

Samuwhale:
Solves #179

### DIFF
--- a/docs/source/faq.rst
+++ b/docs/source/faq.rst
@@ -7,7 +7,7 @@ Here, we've compiled answers to commonly asked questions about metasyn and its d
 
 The project was originally named MetaSynth. However, as we progressed, we discovered that there was already an existing and established audio synthesis software under the same `name <https://uisoftware.com/metasynth/>`_. To avoid potential confusion between these two unrelated projects, we have decided to change the name of our project to **metasyn**. This new name still reflects the package's core goal, of metadata-driven data synthesis. We have also changed the styling of the name to be all lowercase to align with how the package is used in code (e.g. ```import metasyn```).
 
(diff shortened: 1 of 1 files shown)


docs/source/faq.rst
```diff
@@ -3,6 +3,13 @@ FAQ
...
 
+**Why did you change the name from MetaSynth to metasyn?**
+
+The project was originally named MetaSynth. However, as we progressed, we discovered that there was already an existing and established audio synthesis software under the same `name <https://uisoftware.com/metasynth/>`_. To avoid potential confusion between these two unrelated projects, we have decided to change the name of our project to **metasyn**. This new name still reflects the package's core goal, of metadata-driven data synthesis. We have also changed the styling of the name to be all lowercase to align with how the package is used in code (e.g. ```import metasyn```).
+
+It is important to note that despite the name change, metasyn as a project and the package's functionality remain the same. We apologize for any confusion caused by the renaming.
```
qubixes, line 10: Personally, I don't think we need to apologize!

qubixes (approved):
Thanks! apart from the comment, it sounds good to me!

docs/source/faq.rst
Samuwhale: I agree it's definitely not necessary, but I feel like it is a nice gesture that doesn't hurt. But I can remove it if you'd like?

qubixes: Personally, I feel like it is not necessary with an open source project. "We"'re not really an entity like a corporation would be. Of course for now we only have so many contributors, but maybe in the future there will be many more! (hopefully haha)

Samuwhale: That makes sense, changed!

qubixes (approved):
Approved!

Issue #158, opened 2023-08-28
docs: add information on CLI


Samuwhale:
Would be nice to have some information on the CLI usage on the docs.

PR #183, opened 2023-09-29
Added CLI documentation

Samuwhale:
Added documentation on CLI (including docker information), as well as linking to it on other pages.

Solves #158


docs/source/images/pipeline.afphoto~lock~
qubixes, line 1: This probably shouldn't be here?

docs/source/usage/cli.rst
```diff
@@ -0,0 +1,156 @@
+Commandline Interface
+=====================
+Metasyn provides a command line interface (CLI) for accessing core functionality without needing to write any Python code.
+
+The CLI currently has two subcommands. The ``synthesize`` subcommand, which allows you to **generate synthetic data** from any ``GMF file``, and the ``schema`` subcommand, which allows you to **create validation schemas** for a :obj:`MetaFrame<metasyn.metaframe.MetaFrame>`.
```
qubixes, line 5: The validation schema should be for the GMF file, not the MetaFrame I think.

docs/source/usage/cli.rst
```diff
@@ -0,0 +1,156 @@
...
+
+      .. code-block:: console
+
+         docker run -v $(pwd):/wd sodateam/metasyn --help
+
+
+The Metasyn CLI should now be up and running within the Docker container and ready for use!
+
+.. note:: 
+   You can also specify which metasyn version to use in docker, by adding a tag to the docker image. For example, to use version 0.5, you can use the following command:
```
qubixes, line 49: This might be confusing with the name change, since the first metasyn version is 0.6.0.

docs/source/usage/cli.rst
```diff
@@ -0,0 +1,156 @@
...
+
+      .. tab:: Unix or MacOS:
+
+         .. code-block:: console
+
+            docker run -v $(pwd):/wd sodateam/metasyn:v0.5 --help
+
+
+Generating Synthetic Data
+--------------------------
```
qubixes, line 73:
```suggestion
-------------------------
```

docs/source/usage/cli.rst
qubixes: I would just use 0.6 instead.

qubixes (changes requested):
Really nice! I'm still a bit unsure whether the docker installation should have its own place, but I think for now that is okay. Just a small number of points by me!

docs/source/usage/cli.rst
Samuwhale: You're completely right

docs/source/usage/cli.rst
Samuwhale: Yeah the idea was to reflect the fact that older versions can be installed, but you're right, due to there not being an older release, 0.6 is the smarter option.

PR #188, opened 2023-10-04
Progress bar

qubixes:
This adds an optional progress bar if it is estimated to be too slow. I put it at ~5 seconds on my machine, but that can be changed of course.

The PR also makes the FreeText distribution faster by only fitting the first 50 items of the series. It turns out that while benchmarking, this distribution is significantly slower than the regexdistribution. It is also hard to bench, because it gets faster with each run (using the same kernel).

There might be an interesting discussion about how to do the fitting more efficiently. For one, the unique distributions are fit just in the same way as the non-unique ones, so that is basically doing the regexmodel procedure twice.


metasyn/distribution/base.py
Samuwhale, line 1: I'm assuming these estimate formulas are based on performance on your machine - I'm wondering, does it translate well to other machines? I guess it doesn't have to be _that_ accurate since it's just to determine wether or not to place a progress bar, but I'm curious.

Samuwhale:
Looks good though! Is it maybe a nice idea to also allow the user to manually override the automatic progress bar (to let them force enable/disable it)?

metasyn/distribution/base.py
qubixes: Yes, they are. It won't translate to particularly slow machines I suppose, but since most of the calculations are single core, the discrepancy shouldn't be too big.

qubixes:
> Looks good though! Is it maybe a nice idea to also allow the user to manually override the automatic progress bar (to let them force enable/disable it)?

It can already be done, with `progress_bar=True` or `progress_bar=False` for the fit_dataframe method.

Samuwhale:
> > Looks good though! Is it maybe a nice idea to also allow the user to manually override the automatic progress bar (to let them force enable/disable it)?
> 
> 
> 
> It can already be done, with `progress_bar=True` or `progress_bar=False` for the fit_dataframe method.

Ah my bad, I must've missed that. That's nice :)

PR #189, opened 2023-10-09
added an easier way to preview a metaframe's gmf

Samuwhale:
Previously the following syntax had to be used to print the gmf preview of the file:

```
gmf_preview = repr(mf)
print(gmf_preview)
```

A new method `gmf_preview` has been added to allow for easier syntax: `mf.gmf_preview()`


vankesteren (changes requested):
dikke reject

Issue #190, opened 2023-10-09
Add documentation on pretty printing / previewing gmf files 


Samuwhale:
Currently there's no documentation, outside of the notebook examples, on pretty printing / previewing gmf files. Might be nice to add that in the near future.

PR #195, opened 2023-10-13
Added documentation on printing and previewing MetaFrames / GMFs

Samuwhale:
Very small PR, the docs now show that it's possible to print the statistical metadata in a MF, as well as how the `repr` can be used to show the to-be gmf file.

solves #190


qubixes (approved):
Approved!

Issue #178, opened 2023-09-25
Replace flake8 with ruff


qubixes:
I don't know exactly why `_version.py` is found by flake8, but since flake8 doesn't support pyproject.toml, we might want to switch to ruff so we can ignore the file there.

vankesteren:
yes! finally! I had secretly been using ruff already locally ;)

Let's create a good ruff config in pyproject, and then we can also remove pylint I think

Samuwhale:
I'll look into it! :)

PR #199, opened 2023-10-28
Ruff + Tox integration

Samuwhale:
This PR does two things:
- It implements RUFF as a replacement for flake8 (unfortunately replacing other packages is still quite restricted) (config in the [pyproject.to](https://pyproject.to/)ml). It has replaced flake8 in the GitHub workflow. 
- It implements TOX (in the [pyproject.to](https://pyproject.to/)ml), with testing environments, so that one can easily run tests locally (this can be done by using `tox -e [environments]` e.g.: `tox -e ruff,pylint,pydocstyle,mypy,sphinx,pytest,nbval`. To run these tests it is required to install tox into your environment first.
- Theoretically each step in the GH Workflow/Actions could be replaced with TOX, e.g. `tox-e ruff` instead of `ruff metasyn`. Is that something we would want? It might be nice for consistency as it matches how tests can be done locally, but it does also mean that tests will run IN a tox environment IN a GitHub environment, adding an extra layer to the CI.

I am not extremely confident in the way the config is set-up, as it is the first time I have configured these kinds of tools - so I do apologise if the settings and configuration do not fully make sense. 

This solves #178


qubixes (changes requested):
I think for now it's okay to keep the tox out of github actions. Apparently there is a package that connects the two, but it's not that important. I think for now we should see tox as a potential extra, but not core to our dev cycle.

Shouldn't we add ruff, etc to our environment lists in tox, so they get run automatically?

Samuwhale:
> Shouldn't we add ruff, etc to our environment lists in tox, so they get run automatically?

I am a bit confused with what you mean, Ruff is in the tox environment list, right? 

On a side-note, the other comments have been resolved and committed!

pyproject.toml
```diff
@@ -90,3 +90,99 @@ max-args=10
...
+select = ["E", "F", "W", "I", "N"]
+
+[tool.ruff.pep8-naming]
+ignore-names = ["a", "b", "mu"]
+
+[tool.tox]
+legacy_tox_ini = """
+    [tox]
+    env_list =
+        py311
```
qubixes, line 106: I mean this env list. Doesn't it make sense to have also ruff/mypy/etc in there. I don't really know how it works, but I understood that then you can just run the tox command without "-e" to run all the tests.

Samuwhale: You're right! I've fixed this. You can now call them individually with -e, or run them all by simply calling tox.

qubixes (approved):

PR #207, opened 2023-11-06
Updated version number on the docs 

### LINKED ISSUE #206: Docs: the version number is outdated when hovering over the tab on the docs
See example below:

image

This will need to be updated manually with every version release.

Samuwhale:
This fixes #206


qubixes (approved):
Nice! Would be nice if this was automatic, but good to know where to find it. I'll add it to the wiki.

PR #212, opened 2023-11-21
Change from AIC to BIC

### LINKED ISSUE #208: Multinoulli distribution is used too often.
I have noticed with the example dataset that the multinoulli distribution seems to be chosen more often that we would ideally want for the integer distributions. If anyone has an idea how this should be resolved, let's discuss that here.

qubixes:
This seems to help a little with the multinoulli problem, at least "Age" is now correctly identified without using the multinoulli distribution. Fixes #208


vankesteren (approved):
Looks good! If it solves the issue as you say, feel free to merge :)

Samuwhale (approved):
Looks good to me too :)

Samuwhale:
Although small thing - should the docstring in the `information_criterion()` of the `BaseDistribution` be changed?

vankesteren:
Merging so I can use this in my pilot today; will convert comment by @Samuwhale to issue

PR #219, opened 2023-12-01
Constant distributions

### LINKED ISSUE #215: Add constant distribution for all variable types
It's quite important that we add a "constant" distribution for all variable types (i.e. the [Dirac delta function](https://en.wikipedia.org/wiki/Dirac_delta_function) or delta distribution).

The relevant part of the gmf file would look something like
```
'distribution': {
  'implements': 'core.constant',
  'version': '1.0',
  'provenance': 'builtin',
  'class_name': 'ConstantDistribution',
  'parameters': {'value': 'somevalue'}
}
```
Here are some examples of how constant variables are handled currently. 

- Categorical
  ```
  'distribution': {
    'implements': 'core.multinoulli',
    'version': '1.0',
    'provenance': 'builtin',
    'class_name': 'MultinoulliDistribution',
    'parameters': {
      'labels': array(['somevalue'], dtype='<U9'),
      'probs': array([1.])
    }
  }
  ```
- Continuous
  ```
  'distribution': {
    'implements': 'core.uniform',
    'version': '1.0',
    'provenance': 'builtin',
    'class_name': 'UniformDistribution',
    'parameters': {
      'min_val': 1.0, 
      'max_val': 1.0
    }
  }
  ```
- Discrete
  (NB: this is on a branch where I was implementing DiscreteTruncatedNormal but this is fundamentally behaviour we don't want!)
  ```
  'distribution': {
    'implements': 'core.discrete_truncated_normal',
    'version': '1.0',
    'provenance': 'builtin',
    'class_name': 'DiscreteTruncatedNormalDistribution',
    'parameters': {
     'lower_bound': 0.99999999,
     'upper_bound': 1.00000001,
     'mu': 1.0,
     'sigma': 4.999999997368221e-09
    }
  }
  ```
- String
  ```
  'distribution': {
    'implements': 'core.multinoulli',
    'version': '1.0',
    'provenance': 'builtin',
    'class_name': 'MultinoulliDistribution',
    'parameters': {
      'labels': array(['somevalue'], dtype='<U9'),
      'probs': array([1.])
    }
  }
  ```

`

## Choosing ConstantDistribution over current options
All of these approximate the delta function, but we should prefer the ConstantDistribution because it is not an approximation and it is more legible / less ambiguous. To choose delta function over these approximations, we have two options:

- Fast option: pre-check. Are all the values the same -> choose ConstantDistribution. 
- Maintainable / theoretical option: information criterion. Using BIC would also work because the likelihood is infinite at the chosen value for continuous variables, and pmf is 1 at the chosen value for discrete, and we have 0 parameters. However, I don't know if there might be weird interactions with floating point / rounding and such.

### LINKED ISSUE #218: Internal provider naming of distributions with variable types
Currently, we sometimes have a distribution that works for different data types. Motivating example is the uniform distribution. We currently name them (for the distribution provider):

- `core.uniform` (for continuous variables)
- `core.discrete_uniform` (for discrete variables)
- `core.uniform_datetime` (for datetime variables)

Two points: 

1. I don't really like that we have to put the data type in the name while the data type is also a property of the distribution itself, seems to go against DRY
2. There is inconsistency in where we put the data type. 

Let's at least agree on the following proposal which solves point 2: 

__if you include data type, always put it after the distribution name__

So for uniform it will be:

- `core.uniform`
- `core.uniform_discrete`
- `core.uniform_datetime`

(looks more uniform 😉)

An added benefit is that we can put additional modifiers (which according to my point 1 also don't belong in the name but in the properties) in front of the distribution, yielding stuff like:

- `core.unique_uniform`

vankesteren:
Fixes #215 
Fixes #218 
I've been using this basic script to test during development -- still needs to be turned into an actual test:

```python
import polars as pl
from metasyn import MetaFrame

df = pl.DataFrame({
    "floatconst": [1., 1., None, 1., 1.]*100, 
    "intconst": [1, 1, 1, 1, None]*100, 
    "stringconst": ["a"]*500,
    "floatvar": [1.1, 2.3, 1, None, None]*100,
})

mf.synthesize(10)
mf
```

### DIFF
--- a/examples/example_gmf_titanic.json
+++ b/examples/example_gmf_titanic.json
@@ -4,9 +4,9 @@
     "provenance": {
         "created by": {
             "name": "metasyn",
-            "version": "0.6.1.dev32+g871b8ec"
+            "version": "0.6.1.dev35+gf3115b8.d20231130"
         },
-        "creation time": "2023-11-21T13:22:03.439633"
+        "creation time": "2023-11-30T09:40:01.749592"
     },
     "vars": [
         {
@@ -70,13 +70,15 @@
             "dtype": "Int64",
             "prop_missing": 0.19865319865319866,
             "distribution": {
-                "implements": "core.discrete_uniform",
+                "implements": "core.discrete_truncated_normal",
                 "version": "1.0",
                 "provenance": "builtin",
-                "class_name": "DiscreteUniformDistribution",
+                "class_name": "DiscreteTruncatedNormalDistribution",
                 "parameters": {
-                    "low": 0,
-                    "high": 81
+                    "lower_bound": -1e-08,
+                    "upper_bound": 80.00000001,
+                    "mu": 28.403638823278087,
+                    "sigma": 15.862325051407092
                 }
             }
         },
@@ -86,29 +88,15 @@
             "dtype": "Int64",
             "prop_missing": 0.0,
             "distribution": {
-                "implements": "core.multinoulli",
+                "implements": "core.discrete_truncated_normal",
                 "version": "1.0",
                 "provenance": "builtin",
-                "class_name": "MultinoulliDistribution",
+                "class_name": "DiscreteTruncatedNormalDistribution",
                 "parameters": {
-                    "labels": [
-                        0,
-                        1,
-                        2,
-                        3,
-                        4,
-                        5,
-                        6
-                    ],
-                    "probs": [
-                        0.760942760942761,
-                        0.13243546576879914,
-                        0.08978675645342313,
-                        0.0056116722783389455,
-                        0.004489337822671157,
-                        0.0056116722783389455,
-                        0.0011223344556677893
-                    ]
+                    "lower_bound": -1e-08,
+                    "upper_bound": 6.00000001,
+                    "mu": -380.6440825743838,
+                    "sigma": 12.066012048277289
                 }
             }
         },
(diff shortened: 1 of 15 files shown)


qubixes:
Very nice! Did you try and see what happens if you put implements="core.constant" for all distributions?

qubixes:
I have made some big modifications regarding things we have talked about, such as naming. Now distributions don't need to mention the variable type. So if one class implements `core.constant` for `discrete`, another class can implement the same `core.constant` for `continuous`. I have also implemented the constant distribution for `date`, `time` and `datetime`.

examples/example_gmf_titanic.json
```diff
@@ -4,9 +4,9 @@
     "provenance": {
         "created by": {
             "name": "metasyn",
-            "version": "0.6.1.dev35+gf3115b8.d20231130"
+            "version": "0.6.1.dev45+gd3708ea.d20231212"
```
Samuwhale, line 7: When we make a new release, it might be nice to regenerate this file to get a clean version number here. Just a small nitpick though.

Samuwhale (commented):
Based on having read the changes, it looks good. Will do some small testing later though.

examples/example_gmf_titanic.json
qubixes: Sure. But it's going to be changed after the new release anyway..

PR #227, opened 2024-01-04
Refactor metasyn and make CLI more feature complete

qubixes:
Major update for metasyn 1.0.

Breaking/major changes (as far as I know):

- In `MetaFrame.from_dataframe()`, `spec` becomes `var_specs` and changes from a dictionary to a list.
- `MetaFrame`: A new `from_config` classmethod to create a datafree MetaFrame.
- CLI interface is now feature complete using .toml files as input for the configuration.
- CLI: For the `create_meta` method, the input and output parameters are now optional.
- `UniformDistribution` has its parameters changed to `low` and `high` to conform with the discrete version.
- `MetaVar`: The `detect` method is gone and you should now either use the `fit` method or the `__init__` itself.
- `MetaVar`: 
- `Privacy`: Privacy classes can and should now be dynamically loaded using entry points.
- `MetaConfig`: A new configuration object that is used to parse and check toml/dictionaries.
- `VarConfig`: A new variable configuration object.
- `VarConfigAccess`: An accessor for the configuration object that takes into account defaults from the meta_config.
- `DistributionSpec`: A new distribution configuration class that can parse and check configurations.
- `DistributionProviderList`: A new `create` method to allow for datafree creation of distributions.
- `fit_kwargs` is now in the `DistributionSpec`
- `unique` is now in the `DistributionSpec`
Fixes #145

Issue #223, opened 2023-12-11
Update docstrings across metasyn


Samuwhale:
Currently it states that ``to_dict`` must be implemented in derived classes from Base (distribution), but that is not the case. It should be changed to state that ``_param_dict`` , ``default_distribution`` and ``_param_schema`` are required instead.

In general, it would be nice to improve the docstrings - this will also make the automatically generated API on the readthedocs easier to understand.

- [x] Update docstring in Base distribution (fixed in #220)
- [x] Update module level docstrings (fixed in #226)
- [x] Update class level docstrings
- [x] Review and update function level docstrings (I think these are in a relatively okay at the moment, but for 1.1 they should be reverified)

PR #234, opened 2024-01-22
Improved docstrings at a class level

Samuwhale:
This is part of #223 

This is nearly finished, but will wait for #227 to be merged before continuing, due to it having quite a large impact. Note that #227 has been merged into this branch.


Samuwhale:
This PR updates a lot of class level docstrings. Not all have been updated, but I'm opening this PR now because I'd like it merged before continuing, as it is based on quite an outdated branch. Additionally, other issues have priority - so it's nice to have this one merged before continuing work on those.

metasyn/distribution/base.py
```diff
@@ -29,7 +29,7 @@ class BaseDistribution(ABC):
     """Abstract base class to define a distribution.
 
     All distributions should be derived from this class, and should implement the following methods:
-    _fit, draw, _param_dict, _param_schema, default_distribution and __init__.
+    `_fit`, `draw`, `_param_dict`, `_param_schema`, `default_distribution` and `__init__`.
```
qubixes, line 32: Maybe this is useful for sphinx crosslinking?

https://stackoverflow.com/questions/21289806/link-to-class-method-in-python-docstring

metasyn/distribution/base.py
```diff
@@ -229,12 +229,11 @@ def _wrap(cls):
...
 
 class ScipyDistribution(BaseDistribution):
-    """Base class for numerical Scipy distributions.
+    """Base class for numerical distributions using Scipy.
 
     This base class makes it easy to implement new numerical
-    distributions. One could also use this base class for non-scipy
-    distributions, in which case the distribution class should implement
-    logpdf, rvs and fit methods.
+    distributions. It could also be used for non-Scipy distributions,
```
qubixes, line 235:
```suggestion
    distributions. It can also be used for non-Scipy distributions,
```

metasyn/distribution/categorical.py
```diff
@@ -16,13 +16,26 @@
...
+    probs : list of int
+        The probabilities or frequencies of each category. These will be
+        normalized internally.
+
+    Attributes
+    ----------
+    labels : list of str
+        The labels for each category in the distribution.
+    probs : list of int
+        The normalized probabilities of each category.
```
qubixes, line 38: As discussed, this can be removed.

metasyn/distribution/na.py
```diff
@@ -11,7 +11,7 @@
 
 @metadist(implements="core.na", var_type="string")
 class NADistribution(BaseDistribution):
-    """Distribution that will only ever give back NA."""
+    """Distribution that will only consist of NA values (None)."""
```
qubixes, line 14:
```suggestion
    """Distribution that always returns NA values (None)."""
```

metasyn/metaframe.py
```diff
@@ -1,4 +1,4 @@
-"""Module defining the MetaFrame class, used for the conversion of DataFrames to MetaFrames."""
+"""Conversion of DataFrames to MetaFrames."""  # pylint: disable=invalid-name
```
qubixes, line 1: Can you explain a bit why you changed the docstring? To me it seems like there is more in this module then conversion of dataframes.

metasyn/metaframe.py
```diff
@@ -19,24 +19,20 @@
...
+
+    A MetaFrame represents a metadata frame, which is a structure that holds
+    statistical metadata about a dataset. The data contained in a MetaFrame
+    follows the Generative Metadata Format (GMF).
+    The metadata is contained in a collection of MetaVar objects,
+    with each MetaVar representing a column (variable).
+
+    A MetaFrame can easily be created using the 'fit_dataframe()' method,
+    which takes a Polars DataFrame and fits a MetaFrame to it.
     """
```
qubixes, line 36: Where have the parameters gone?

metasyn/provider.py
```diff
@@ -61,9 +61,13 @@
...
 class BaseDistributionProvider(ABC):
-    """Class that encapsulates a set of distributions.
+    """Base class for all distribution providers.
 
-    It has a property {var_type}_distributions for every var_type.
+    A distribution provider is a class that provides a set of distributions
+    that can be used by metasyn to generate synthetic data.
+    This class acts as a base class for creating specific distribution
+    providers. It encapsulates a set of distributions and has a property
+    {var_type}_distributions for every var_type.
```
qubixes, line 70: Not anymore! Currently they are all in one list. Maybe some documentation on the legacy distributions would be nice as well.

metasyn/var.py
```diff
@@ -45,6 +45,11 @@ class MetaVar():
         back.
     description:
         User-provided description of the variable.
+
+    Attributes
+    ----------
+    dtype : str
+        The data type of the original values. Default is "unknown".
```
qubixes, line 52: This can probably go now as well. The default should be part of the parameters though.

qubixes (changes requested):
Thanks for looking through the doc strings. Seems like there are some out-of-date items on there as well!

metasyn/metaframe.py
Samuwhale: I think I was actually sloppy here, and forgot that I was still working on this part and never looked back to it. Updated it again.

metasyn/provider.py
Samuwhale: Changed it, is this better?

docs/source/api/metasyn.distribution.rst
```diff
@@ -1,13 +1,8 @@
...
+============================
 
-.. automodule:: metasyn.distribution
-   :members:
-   :undoc-members:
-   :show-inheritance:
-
-Submodules
-----------
+The :mod:`metasyn.distribution` package consists of several distribution modules.
```
qubixes, line 4: Isn't this supposed to go in the docstrings?

metasyn/distribution/base.py
```diff
@@ -28,8 +29,14 @@ class attributes of a distribution.
 class BaseDistribution(ABC):
     """Abstract base class to define a distribution.
 
-    All distributions should be derived from this class, and should implement the following methods:
-    `_fit`, `draw`, `_param_dict`, `_param_schema`, `default_distribution` and `__init__`.
+    All distributions should be derived from this class, and should implement
+    the following methods:
+    :meth:`~metasyn.distribution.base.BaseDistribution._fit`,
```
qubixes, line 34: From what I read, you can also do relative paths, maybe that will be slightly easier to read at least in the code.

metasyn/metaframe.py
qubixes: Good!

qubixes (changes requested):
Overall looks good to me, just some small comments, and I think it can be merged pretty quickly.

metasyn/distribution/base.py
Samuwhale: I actually think this reads nicer than:

```
    :meth:`~_fit`,
    :meth:`~draw`,
    :meth:`~_param_dict`,
    :meth:`~_param_schema`,
    :meth:`~default_distribution`
    and ``__init__``.
    
  ```
  
  In this context, do you have a strong preference against this?

File: docs/source/api/metasyn.distribution.rst (hunk 9 above)
Comment (Samuwhale) [reply docs/source/api/metasyn.distribution.rst:4]: Nope, because of the way the API is structured now it wouldn't show up.

File: metasyn/distribution/base.py (hunk 10 above)
Comment (qubixes) [reply metasyn/distribution/base.py:34]: I do think it's a bit nicer yeah, might be able to fit them on the same line. Btw, I saw that some of these distributions are starting with an underscore and don't show up in the docs. Maybe it would be good for this class to show at least the abstract underscore classes and improve the docstrings on the a bit, because they're really short right now.

File: docs/source/api/metasyn.distribution.rst (hunk 9 above)
Comment (qubixes) [reply docs/source/api/metasyn.distribution.rst:4]: Can you explain why we can't have the docstring in the metasyn.distribution.__init__ can't be shown in its place there?

Comment (Samuwhale) [reply docs/source/api/metasyn.distribution.rst:4]: No okay, technically it is possible. But I have restructured the API page in a way that the modules are listed individually, instead of being listed by autodocumenting the distribution package, as such it wouldn't show. 

I did this because the page used to list the modules in both ways (through the distribution package and individually), and was very long, redundant and mostly confusing as a result. And because it gives flexibility, as it now allows me to change the order of the modules (putting 'base' first), and add extra text or formatting as need be. It also allows for neater sidebar formatting.

That said I realise this is not as easy to maintain going forward, so I will change it to autodoc the whole distribution package, and I'll move the text to the module docstring while I'm at it.

File: metasyn/distribution/base.py (hunk 10 above)
Comment (Samuwhale) [reply metasyn/distribution/base.py:34]: Yeah those disabled links and lacking documentation is bothering me too. I would love to show it, but I am a bit confused. How do you mean showing them?

File: docs/source/api/metasyn.distribution.rst (hunk 9 above)
Comment (Samuwhale) [reply docs/source/api/metasyn.distribution.rst:4]: Never mind, I have looked at it - and the overall flexibility and better organizing of the page is just a lot better when listing all the modules individually. I think the slight increase in maintainability effort is worth it (it also significantly impacts sidebar navigation), so it will have to stay as a separate string at the top of the doc.

Comment (qubixes) [reply docs/source/api/metasyn.distribution.rst]: Okay, if it works out better, I'm okay with it.

File: metasyn/distribution/base.py (hunk 10 above)
Comment (qubixes) [reply metasyn/distribution/base.py]: Yes, they are showing now in the documentation, thanks!

File: docs/source/api/metasyn.distribution.rst [hunk 11]
```diff
@@ -1,70 +1,97 @@
...
 .. automodule:: metasyn.distribution.faker
    :members:
    :undoc-members:
+   :inherited-members:
+   :private-members:
    :show-inheritance:
 
-metasyn.distribution.regex module
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Na module
```
qubixes, line 77:
```suggestion
NA module
```
I think it's generally both capitalized.

qubixes (approved):
Only one tiny comment, otherwise I approve!

PR #236, opened 2024-01-26
Update LICENSE

Samuwhale:
Updates the license date

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2023 SoDa
+Copyright (c) 2024 SoDa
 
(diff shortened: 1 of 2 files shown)


vankesteren:
I always wondered if this is something we need to do? Or is it fine to keep the original date?

Samuwhale:
From what I gather it's fine to keep it as is, but it is common practice to keep them updated. That said, I am definitely not knowledgeable on copyright laws, so I'm unsure about specifics.

qubixes (approved):
Often you see with big software a range, like 2022-2024, but I think updating it each year is fine as well.

PR #239, opened 2024-01-31
Homogenize distribution parameters

### LINKED ISSUE #228: Check the names of all parameters of distributions.
I think all distributions with the same name should have the same parameter names as well. Also, it would be good to make sure we agree on all parameter names before we make 1.0.

vankesteren:
fixes #228

### DIFF
--- a/examples/basic_example.py
+++ b/examples/basic_example.py
@@ -27,11 +27,6 @@ specs = [
     VarConfig(name="B", dist_spec=DistributionSpec(unique=True))
 ]
 
-# specs = {
-#     "ID": {"unique": True},
-#     "B": {"unique": False}
-# }
-
 # create MetaFrame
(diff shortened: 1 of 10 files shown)


examples/basic_example.py
```diff
@@ -20,13 +22,18 @@
...
-}
+specs = [
+    VarConfig(name="ID", dist_spec=DistributionSpec(unique=True)),
+    VarConfig(name="B", dist_spec=DistributionSpec(unique=True))
+]
+
+# specs = {
+#     "ID": {"unique": True},
+#     "B": {"unique": False}
+# }
```
qubixes, line 33:
```suggestion
```

qubixes (approved):
Thanks! I didn't find anything out of the ordinary, mostly just find and replace.

PR #244, opened 2024-02-01
Change consecutive parameter in unique_key to bool

### LINKED ISSUE #237: UniqueKeyDistribution consecutive param bool?
In my opinion it would be better to set the "consecutive" parameter of the UniqueKeyDistribution to boolean. Do you think this will break things @qubixes?

qubixes:
Fixes #237

### DIFF
--- a/metasyn/distribution/discrete.py
+++ b/metasyn/distribution/discrete.py
@@ -136,7 +136,7 @@ class UniqueKeyDistribution(ScipyDistribution):
         1 if keys are consecutive and increasing, 0 otherwise.
     """
 
-    def __init__(self, lower: int, consecutive: int):
+    def __init__(self, lower: int, consecutive: bool):
         self.par = {"lower": lower, "consecutive": consecutive}
         self.last_key = lower - 1
         self.key_set: Set[int] = set()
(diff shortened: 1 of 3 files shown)


metasyn/distribution/discrete.py
```diff
@@ -146,8 +146,8 @@ def _fit(cls, values):
         lower = values.min()
         high = values.max() + 1
         if len(values) == high-lower and np.all(values.to_numpy() == np.arange(lower, high)):
```
vankesteren, line 148: Maybe this comparison is slightly inefficient?

vankesteren (approved):

metasyn/distribution/discrete.py
qubixes: How would you prefer to do it?

vankesteren: not sure, in R I'd do `all(diff(values) == 1)` then we don't need to instantiate a completely new vector which might be a million values. But it's a super small comment which you could ignore as well

qubixes: I don't actually know if `np.arange` will instantiate a new vector. Let's for now keep it like it is. I don't think this will really ever be a bottleneck, and if so we can revise then.

PR #246, opened 2024-02-02
Fix a small cli doc error

vankesteren:


qubixes (approved):
Seems like a good change to me!

PR #248, opened 2024-02-19
Add Python 3.12 as possible option for install

qubixes:


vankesteren:
nice, seems like there are no problems at all!

PR #254, opened 2024-02-25
Fix categorical type annotation

vankesteren:


qubixes (approved):
Yes, that seems like an oversight indeed!

PR #256, opened 2024-02-26
Demo datasets

### LINKED ISSUE #252: Fun dataset for an example someday 🚀
https://www.kaggle.com/c/spaceship-titanic

It's cc-by as well so we might even include it as a polars data frame in the package for easy examples

vankesteren:
fixes #252

### DIFF
--- a/examples/basic_example.py
+++ b/examples/basic_example.py
@@ -1,30 +1,14 @@
-import polars as pl
-
-from metasyn import MetaFrame
+from metasyn import MetaFrame, demo_data
 from metasyn.config import VarConfig
 from metasyn.util import DistributionSpec
 
 # example dataframe from polars website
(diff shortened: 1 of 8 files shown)


qubixes (changes requested):
Thanks for adding the new dataset. I do worry a little about using demo_data instead of demo_file. Especially, because during the workshop my guinea pigs told me that they would like to see more explanation of polars in the README, not less. Of course, the example/tutorial itself becomes smoother, but the experience for the user after using the tutorial on their own data might become less smooth.

For the naming, I would prefer demo_dataframe, since it will give the user much more information on what's in there.

qubixes (approved):
Thanks! It's definitely nice to have the demo_dataframe functionality as well!

PR #259, opened 2024-02-27
Fix to_dict method of BasePrivacy

### LINKED ISSUE #253: Bug: calling metasyn with instance of BasePrivacy as privacy argument
reprex: 

```py
from metasyn import MetaFrame
from metasyn.privacy import BasicPrivacy
import polars as pl

mf = MetaFrame.fit_dataframe(
    df=pl.DataFrame({"ok": ["test", "test", "test"]*11}), 
    privacy=BasicPrivacy()
)
```

bug is because `BasicPrivacy().to_dict()` returns 

```py
{'type': 'none', 'parameters': {}}
```

rather than

```py
{'name': 'none', 'parameters': {}}
```

This is one small bug in a whole host of problems I found while trying out disclosure-control on the YOUth data. We need to do a hackathon to get that working.

qubixes:
Fixes #253

### DIFF
--- a/metasyn/privacy.py
+++ b/metasyn/privacy.py
@@ -29,7 +29,7 @@ class BasePrivacy(ABC):
     def to_dict(self) -> dict:
         """Create a dictionary that gives the privacy type, and parameters."""
         return {
-            "type": self.name,
+            "name": self.name,
             "parameters": {},
         }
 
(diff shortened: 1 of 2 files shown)


vankesteren (approved):

vankesteren:
Just need to adjust the test and then feel free to merge!

PR #260, opened 2024-02-27
Fix boolean types to be supported as categorical

### LINKED ISSUE #255: Support bool datatype
We should map bool to categorical!

traceback:
```py
File ~\surfdrive\SoDa\projects\synthetic_data\metasyn\metasyn\var.py:104, in MetaVar.get_var_type(series)
    102     return convert_dict[polars_dtype]
    103 except KeyError as exc:
--> 104     raise ValueError(
    105         f"Unsupported polars type '{polars_dtype}'") from exc

ValueError: Unsupported polars type 'bool'
```

qubixes:
Fixes #255 

I don't know if we should have an actual boolean type, but this seems to work fine for now.

### DIFF
--- a/metasyn/var.py
+++ b/metasyn/var.py
@@ -96,6 +96,7 @@ class MetaVar():
             "time": "time",
             "str": "string",
             "categorical": "categorical",
+            "bool": "categorical",
             "NoneType": "continuous",
         }
         try:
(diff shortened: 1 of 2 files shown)


vankesteren:
This would have been exactly my approach, nice and clean. Are we testing that the synthesized data is of type "bool" as well?

qubixes:
@vankesteren I have now added a test for that.

PR #267, opened 2024-02-29
Remove notebook checking for osx

### LINKED ISSUE #263: macOS ci build often fails on nbval
See for example here: https://github.com/sodascience/metasyn/actions/runs/8066127477/job/22033475174

A rerun often fixes it. But it takes 37 minutes off our budget every time so that's very wasteful! We should either fix this or turn off macOS builds for now

qubixes:
This will skip the test to check the notebooks on mac OS, which can fail (still). Fixes #263

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -54,6 +54,6 @@ jobs:
       run: |
         pytest tests
     - name: Check notebook output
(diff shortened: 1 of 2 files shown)


vankesteren (approved):
Yes, cool. It's very opaque why this happens though! Maybe at some point we can retry.

PR #272, opened 2024-03-05
Add creation settings

### LINKED ISSUE #258: Add provenance information for each variable, including privacy parameters
Currently there is some provenance information in the distribution section. However, this does not include any information how it was created, just by whom.

qubixes:
Fixes #258 

This PR tries to add some more information about how the variable was created. It probably needs a little bit more testing.

### DIFF
--- a/metasyn/__main__.py
+++ b/metasyn/__main__.py
@@ -83,7 +83,7 @@ def main() -> None:
         sys.exit(1)
 
 
-def create_metadata():
+def create_metadata() -> None:
     """Program to create and export metadata from a DataFrame to a GMF file (.json)."""
     parser = argparse.ArgumentParser(
         prog="metasyn create-meta",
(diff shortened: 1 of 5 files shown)


Samuwhale (approved):
Looks good to me!

vankesteren (commented):
Looks good, the logic looks done! One question: does "creation settings" still cover the meaning? Would "fit parameters", fitting config, estimation settings be better?

qubixes:
> Looks good, the logic looks done! One question: does "creation settings" still cover the meaning? Would "fit parameters", fitting config, estimation settings be better?

I think those are okay as well. However, I think the point is that some distributions are not estimated / fitted, but simply defined. If you have a suggestion that also covers that use case, then that might be better than creation_settings.

vankesteren:
After thinking more, I'd go for either:
- fit_method
- creation_method

I still think "fit" works, as the user is also "fitting", just not doing ML estimation

qubixes:
> After thinking more, I'd go for either:
> 
>     * fit_method
> 
>     * creation_method
> 
> 
> I still think "fit" works, as the user is also "fitting", just not doing ML estimation
> 
> You can choose and merge!!

Okay, creation_method it is then!

PR #283, opened 2024-03-25
Better errors

### LINKED ISSUE #265: Improve error messages and warnings
In each of our steps (fit_dataframe, export, from_json, synthesize) it would be nice to think a bit more about friendly and helpful error messages. Especially in the first one, it's very important to attend users to the different options. Also, we used to have this warning for unique values which is now gone?

I feel like this is an area where we can look at the tidyverse once again. Let's think about what can go wrong for the user in each of these steps and come up with nice error messages!

Samuwhale:
Hi! This PR fixes #265. It doesn't fully handle all the errors in the list yet, but I thought it would be nice to create a draft PR already to get some opinions on the current work. Let me know any thoughts!

### DIFF
--- a/metasyn/config.py
+++ b/metasyn/config.py
@@ -97,17 +97,18 @@ class MetaConfig():
         meta_config:
             A fully initialized MetaConfig instance.
         """
-        if Path(config_fp).suffix == '':
+        try:
+            with open(config_fp, "rb") as handle:
+                config_dict = tomllib.load(handle)
+        except FileNotFoundError as fnf_error:
             raise FileNotFoundError(f"It appears '{config_fp}' is not a valid filepath."
                                     f" Please provide a path to a .toml file to load a MetaConfig"
-                                    f" from.")
-        if Path(config_fp).suffix != '.toml':
+                                    f" from.") from fnf_error
+        except ValueError as value_error:
             raise ValueError(f"It appears '{Path(config_fp).name}' is a"
                              f" '{Path(config_fp).suffix}' file."
-                             f" To load a MetaConfig, provide it as a .toml file.")
+                             f" To load a MetaConfig, provide it as a .toml file.") from value_error
 
-        with open(config_fp, "rb") as handle:
-            config_dict = tomllib.load(handle)
         var_list = config_dict.pop("var", [])
         n_rows = config_dict.pop("n_rows", None)
         dist_providers = config_dict.pop("dist_providers", ["builtin"])
--- a/metasyn/config.py
+++ b/metasyn/config.py
@@ -97,6 +97,10 @@ class MetaConfig():
         meta_config:
             A fully initialized MetaConfig instance.
         """
+        if not str(config_fp).endswith('.toml'):
+            raise ValueError(f"It appears '{config_fp}' is not a .toml file."
+                             f" To load a MetaConfig, provide it as a .toml file.")
+
         with open(config_fp, "rb") as handle:
             config_dict = tomllib.load(handle)
         var_list = config_dict.pop("var", [])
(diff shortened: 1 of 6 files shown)


metasyn/metaframe.py
```diff
@@ -245,6 +248,8 @@ def export(self, fp: Optional[Union[pathlib.Path, str]],
         if fp is None:
             print(json.dumps(self_dict, indent=4))
         else:
+            if pathlib.Path(fp).is_file():
```
qubixes, line 251: This can go

metasyn/privacy.py
```diff
@@ -90,6 +90,15 @@ def get_privacy(name: str, parameters: Optional[dict] = None) -> BasePrivacy:
         if name == entry.name:
             return entry.load()(**parameters)
     privacy_names = [entry.name for entry in entry_points(group="metasyn.privacy")]
+
```
qubixes, line 93: Please remove

qubixes (changes requested):

metasyn/provider.py
```diff
@@ -498,8 +505,8 @@ def _get_all_provider_list() -> list[BaseDistributionProvider]:
 
 def get_distribution_provider(
         provider: Union[str, type[BaseDistributionProvider],
-                        BaseDistributionProvider] = "builtin"
-        ) -> BaseDistributionProvider:
+        BaseDistributionProvider] = "builtin"
```
qubixes, line 508: Can you revert this? Now it is wrongly aligned.

metasyn/config.py
```diff
@@ -97,6 +97,15 @@ def from_toml(cls, config_fp: Union[str, Path]) -> MetaConfig:
         meta_config:
             A fully initialized MetaConfig instance.
         """
+        if Path(config_fp).suffix == '':
+            raise FileNotFoundError(f"It appears '{config_fp}' is not a valid filepath."
+                                    f" Please provide a path to a .toml file to load a MetaConfig"
+                                    f" from.")
```
qubixes, line 103: I think we should do it the other way around: try to load the file and if it doesn't load, show the errors. In the try/catch you should catch the error and reraise it. So if you get a FileNotFoundError, reraise from there. Otherwise you probably get some kind of Parse error. In that case you can check for the suffix and give a hint depending on that.

qubixes (changes requested):
Thanks! Some small changes still needed (also to pass the test). Btw, it seems we still call the config files config.ini in the tests, so maybe that can be changed as well.

metasyn/config.py
Samuwhale: I've changed it - is this what you had in mind?

qubixes: Yes, thanks! The only thing I would split is that there can be parse errors because the .toml file is broken or a non-toml file is supplied. Perhaps we can split the error message between the two with the suffix test that you made before?

Samuwhale: Done! Feel free to suggest a change to the phrasing.

metasyn/provider.py
```diff
@@ -496,10 +503,8 @@ def _get_all_provider_list() -> list[BaseDistributionProvider]:
     return [p.load()() for p in _get_all_providers().values()]
 
 
-def get_distribution_provider(
-        provider: Union[str, type[BaseDistributionProvider],
-                        BaseDistributionProvider] = "builtin"
-        ) -> BaseDistributionProvider:
+def get_distribution_provider(provider: Union[str, type[
+    BaseDistributionProvider], BaseDistributionProvider] = "builtin") -> BaseDistributionProvider:
```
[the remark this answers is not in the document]
Samuwhale: I changed it, let me know if this is what you intended!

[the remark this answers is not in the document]
Samuwhale: It's ready to merge if the indentation is okay

qubixes:
I'm in favor of a merge.

Issue #291, opened 2024-04-17
Command-Line Interface docs out of date


LindaNab:
Hi `metasyn`-developers,

I'm playing around with `metasyn` (it's awesome!) and while using the CLI functionality to check out some of the basic commands I got across an error message I could not resolve. I think the documentation on readthedocs might be a bit out of sync with the current version of `metasyn`? 

In specific, I was hoping to create a `.json` file using a user defined `.toml` file (without an input CSV file). According to your documentation under this [section](https://metasyn.readthedocs.io/en/latest/usage/cli.html#creating-generative-metadata), an example `.toml` file is:
```
n_rows = 100


[[var]]

name = "PassengerId"
data_free = true
prop_missing = 0.0
description = "ID of the unfortunate passenger."
var_type = "discrete"
distribution = {implements = "core.unique_key", unique = true, parameters = {consecutive = 1, lower = 0}}
```

When I subsequently try to run:
```
metasyn create-meta --config my_config.toml --output my_gmf.json
```
I get the following error message: (truncated)
```
jsonschema.exceptions.ValidationError: {'implements': 'core.unique_key', 'version': '1.0', 'provenance': 'builtin', 'class_name': 'UniqueKeyDistribution', 'unique': True, 'parameters': {'lower': 1, 'consecutive': 1}} is not valid under any of the given schemas
Failed validating 'anyOf' in schema['properties']['vars']['items']['properties']['distribution']:
[...]
```
Please note that I changed `low` as per your documentation to `lower` in my `.toml` as I figured `low` isn't a parameter of the `core.unique_key` distribution while `lower` is. I also added the argument `--output` before my output file in the CLI command. The argument `--output` was not listed in your documentation but without `--output` the output of `metasyn create-meta` is printed in the terminal and not saved to the output file, but please correct me if I'm wrong...

I did manage to implement the `core.truncated_normal` distribution but can't figure how to make `core.unique_key` work... I tried several combinations of the parameters `lower` and `consecutive` but according to the error message their type is a `integer` and `boolean` respectively, so I can't really make sense of the error. Help would be much appreciated!

Best wishes,
Linda Nab

qubixes:
Hi @LindaNab,

Thanks for the opening an issue, and you're right the documentation is indeed out of date!

I can reproduce the error with the latest version of metasyn (main branch). I can fix it on my end by setting `consecutive = true`. Does the example work for you then?

[...]

PR #292, opened 2024-04-18
Remodel API docs with inspiration from Scipy

qubixes:
I have looked at our API and currently it's not very readable (imo), also with the eye on #291. This is a draft PR, because more organization needs to be done. For inspiration I took:

https://github.com/scipy/scipy/blob/main/scipy/spatial/distance.py

I think the API reference can be a little bit more guided, and I think the autosummary makes a lot of sense in this context. I would like some feedback on the new arrangement before I continue to work on it.

On a related note: we should probably move all distributions with the same type into that module. Now there are for example four places where we have string distributions (faker.py/constant.py/categorical.py/regex.py). The Multinoulli distribution is special of course, but for the rest I think we're better off putting them all in the same module (and if it's getting to long make a package out of it).


qubixes:
I can't get it to pass the sphinx test without warning. I'm tempted to just remove the warning on error part of the CI. Anyone has an idea on how to fix it?

vankesteren:
feel free to merge!

Samuwhale:
I did not have time to review this in time earlier, but just wanted to say nice work! Looks a lot more organized.

PR #296, opened 2024-06-07
Fix defaults so plugins don't need manual fitkwarg

qubixes:
While technically a bug of the metasyn-disclosure control (See https://github.com/sodascience/metasyn-disclosure-control/issues/27), this is easier fixed in the main metasyn library.

I haven't had the time to test this fix yet.


qubixes:
Adding some fixes to this PR so that the tests succeed again.

qubixes:
The CI gives errors, because of issues related to the disclosure control plugin, but should pass once these are fixed.

vankesteren (approved):

PR #298, opened 2024-07-02
Set information criterion for NA distribution high

qubixes:
Not setting a high IC for the NA distribution causes the NA distribution to be selected for the disclosure control plugin (since there is not regex model). This PR should fix that.

There might be some more discussion to be had on the relative IC's for the static type of distributions, but I think for now, we should never select the NA distribution unless they're all NA's (or no other distributions are available).


qubixes:
@vankesteren Shall I just merge it?

vankesteren:
Yes!

PR #299, opened 2024-07-11
Move polars to version 1, fix test errors, rerun examples

vankesteren:


qubixes (approved):
Thanks, I suppose with too many samples, even though the uniform distribution should still be favored, it will be closer and closer to choosing the multinomial distribution?

vankesteren:
Exactly!

PR #311, opened 2024-08-22
Add VarSpec to remove the warnings from the examples

qubixes:
This is one way to do it. Another would be to get a demo file without a unique column. It might be helpful to show the VarSpec off anyway?

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ In Python code this happens as follows:
 
 ```python
 import polars as pl
-from metasyn import MetaFrame, demo_file
+from metasyn import MetaFrame, demo_file, VarSpec
 
 # Get the csv file path for built-in demo dataset
 csv_path = demo_file("fruit")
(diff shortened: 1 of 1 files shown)


vankesteren (changes requested):
i prefer the other solution you mentioned. the initial example should be as simple as possible

qubixes:
Shall I adjust the fruit dataset then?

vankesteren:
hmm, does spaceship give an error? otherwise we can do that one. Fruit is take. directly from the polars website so let's keep it the same as there.

we can also just create / find a new example dataset like "questionnaire" or so, with demographics and some questionnaire likely scale answers and an open question

qubixes:
Spaceship does not give an error, however the image in the readme is about fruit... The spaceship dataset is quite wide to show easily like that.

vankesteren:
I've reconsidered. I now think that receiving a warning is not so bad in the main example. Maybe we can leave it as-is and update/improve the warning note instead?

qubixes:
Sure, I can think about that. It does sound a little bit like something has gone wrong, which is not necessarily the case.

vankesteren:
Is there something like a "note" in python rather than a "warning"?

qubixes:
We could use the logging module in the standard library. This might be redirected to stdout and not stderr though, which I think is the case of the warnings module?

Issue #307, opened 2024-08-19
User test feedback


vankesteren:
@Samuwhale (thanks!!) did user tests with several users unfamiliar with metasyn, and this is the most important feedback that came out of those tests.

- [x] #315
- [x] The most common problem: (all tested) users expected synthetic data generation parameters (e.g. Faker names, Unique, etc.) to be under ‘Generating Synthetic Data’ on the readme, as opposed to ‘Generating MetaFrames’. People only found the relevant page after using the search function, after already having looked at multiple wrong pages. -> Adding extra information on this page guiding users to the appropriate one is crucial. 
- [x] One user found VarSpec to be a confusing name, and suggested it being called something along the lines of ‘’extra generation parameters for variables’’
- [x] Python 3.12 does not allow straight up installation, and enforces users to install in a virtual environment -> this COULD potentially be mentioned as a small not in the install guide (e.g. “when using python 3.12 do …)
- [x] #328
- [x] The documentation redirect on github is broken (to learn more about how `metasyn` works, go to detailed overview in our documentation)
- [x] #309
- [x] One user tried using pandas instead of polars, and was confused about it not working. Maybe this should be made clearer somewhere - or a clear warning should be given about this when it occurs?

qubixes:
I think we can close this issue.

PR #313, opened 2024-08-29
Add survey demo dataset

### LINKED ISSUE #314: Dtype conversion issue
@qubixes you had excellent foresight

https://github.com/sodascience/metasyn/blob/c1029053309c5c8646814f933fca036a5c6bbcd5/metasyn/var.py#L78-L79

it broke. I fixed it for the categorical dtypes in #313, but we might want to do something a bit more stable / custom like have a direct lookup table from polars dtypes (e.g., `pl.Categorical`) to metasyn.

vankesteren:
Data from the European Social Survey, selected. Free to use for non-commercial purposes as per the "data access" section in https://ess.sikt.no/en/study/412db4fe-c77a-4e98-8ea4-6c19007f551b/86.

This will be useful in potentially fixing the first point in #307  

Fixes #314


vankesteren:
Thanks, fixed and will merge when tests pass

Issue #282, opened 2024-03-20
Separate our GH actions


vankesteren:
Currently, we have 1 GH action that does everything. However, it might be nice to separate it out into several actions, together with our changes here: 
- #278 

For example, not run the full matrix all the time, not run the full documentation all the time, etc.

vankesteren:
- Linting can be done on a single OS
- Plug-in testing should be done separately
- We should separate our optional-dependencies in pyproject.toml further to help this

qubixes:
The one advantage of doing everything in one go is that is saves installation time. Let's see how much more it costs to separate the testing.

PR #316, opened 2024-08-30
Simplify CI & development experience

vankesteren:
Here is my proposal for a simpler development and CI experience. 

- Only use ruff for linting, which is now in a really good state to replace pydocstyle, flake8, pylint, etc
- Improve CI actions:
  - Scheduled CI on main and development every week
  - Standard CI now runs in parallel (lint/check, test, and doc build) through jobs
  - publish to pypi and dockerhub when a new release is created
- Separate optional dependencies into `test`, `check`, `extra`, `docs`, and `dev` (which contains everything)

Dev process will be: 
- (fork and) clone the repo
- `pip install -e .[dev]`
- create changes
- `ruff check`
- `mypy metasyn`
- `pytest tests`
- `pytest --nbval-lax examples`
- and you're good to commit!


vankesteren:
I've added pylint rules back in through ruff and removed mentions of pylint

vankesteren:
I'll add more pylint rules next

vankesteren:
Added all pylint rules supported by ruff. Also added pydocstyle rules (and excluded some particularly ignored ones)

vankesteren:
Fixes #282

vankesteren:
After the last commit, I double-checked that the documentation still builds on rtd:

https://metasyn.readthedocs.io/en/update-workflows/

qubixes (changes requested):
Okay for everything except for the formatting changes. Personally, I don't think they are an improvement, but I can be convinced!

vankesteren:
Ah yes, you are right, will fix

vankesteren:
@qubixes are you happy like this?

metasyn/varspec.py
```diff
@@ -64,6 +64,7 @@ def parse(cls, dist_spec: Optional[Union[dict, type[BaseDistribution], BaseDistr
         ------
         TypeError
             If the input has the wrong type and cannot be parsed.
+
```
qubixes, line 67: I missed this variation, but this extra line also seems a little bit extraneous? If I look at PEP 257 it doesn't seem to mention anything like that?

vankesteren: Your're right, good catch, this was also introduced when I hadn't yet configured ruff properly

qubixes:
Okay, approved then, thanks!

PR #355, opened 2025-01-10
Add rudimentary file handling support.

qubixes:
This PR is now in a reasonable state. It needs more tests though to pass the threshold of 95% coverage.
I have updated the documentation, tutorials, docstrings, etc.

The problem I have run into, when trying to use SPSS (.sav) files is that there is a lot of information stored outside of what can be captured in the dataframe. For example column labels, formats (how many decimal points), file labels, tables for categories, etc. When you synthesize the data from the resulting GMF file (which does not include this information), you get an SPSS file that is not quite comparable to the original, which partly defeats the purpose of metasyn.

Another example with CSV files could be that the delimiter could be different from the original versus the synthetic data.

This information is now stored in the GMF file.

There are two file handlers implemented: .sav and .csv, so both of those can now be read and written back as a synthetic file.

New API:

```python
from metasyn import demo_file, MetaFrame
import metasyn as ms


fp = demo_file("titanic")
df, file_reader = ms.read_csv(fp)
mf = MetaFrame.fit_dataframe(df, file_reader=file_reader)
mf.save("test.json")
mf = MetaFrame.load("test.json")
mf.write_synthetic(file_name="synthetic_titanic.csv")
```

Example CLI:
```bash
metasyn create-meta "$INPUT_FILE" -o /tmp/gmf.json
metasyn synthesize /tmp/gmf.json -o "$OUTPUT_FILE"
```


vankesteren:
First overall request for change:

I think the interface should not interfere with our current workflow. So how you've implemented it now:

```py
import polars as pl
from metasyn import MetaFrame, demo_file
from metasyn.filehandler import get_file_handler
fp = demo_file("spaceship")

# Loading convolves df with file meta
df, fmt = get_file_handler(fp) # (also this function name is not great)

# Now we need to do loads of additional stuff with df here to change schema!!!
df = df.with_columns(
    pl.col("HomePlanet").cast(pl.Categorical),
    pl.col("CryoSleep").cast(pl.Categorical),
    pl.col("VIP").cast(pl.Categorical),
    pl.col("Destination").cast(pl.Categorical),
    pl.col("Transported").cast(pl.Categorical),
)

# As usual, nice

# This part is nice too
mf.file_format = fmt

# nice
mf.write_synthetic("test.csv")
```

I think a better interface would keep everything as-is and only introduce an optional part:

```py
import polars as pl
from metasyn import MetaFrame, demo_file
from metasyn.fileformats import read_file_format
fp = demo_file("spaceship")

# read data and create mf as always
schema = {
    "HomePlanet": pl.Categorical,
    "CryoSleep": pl.Categorical,
    "VIP": pl.Categorical,
    "Destination": pl.Categorical,
    "Transported": pl.Categorical,
}
df = pl.read_csv(fp, try_parse_dates=True, schema_overrides = schema)

# Add file format, completely optional!!
mf.file_format = read_file_format(fp) # completely optional!

### PROBLEM: for files like sav we need to check that the columns are there? ###

# this should error before synthesizing if there is no file format (it currently does not)
mf.write_synthetic("test.sav")
```

metasyn/filehandler.py
vankesteren, line 1: rename to something like fileformats.py, see main comment

metasyn/filehandler.py
```diff
@@ -0,0 +1,191 @@
...
+@filehandler
+class SavFileHandler(BaseFileHandler):
+    name = "spss-sav"
+    extensions = [".sav", ".zsav"]
+
+    def read_dataset(self, fp):
+        df, _prs_metadata = self._get_df_metadata(fp)
+        return df
+        # df, prs_metadata = pyreadstat.read_sav(fp, apply_value_formats=True)
+        # return pl.DataFrame(df)
```
vankesteren, line 62: this should be out of scope

metasyn/filehandler.py
```diff
@@ -0,0 +1,191 @@
...
+def filehandler(*args):
+    """Register a dataset so that it can be found by name."""
+
+    def _wrap(cls):
+        _AVAILABLE_FILE_HANDLERS[cls.name] = cls
+        return cls
+
+    return _wrap(*args)
+
+class BaseFileHandler(ABC):
```
vankesteren, line 21: I'm happy with filehandler as internal name, but any user-facing function and message should be clearer w.r.t. functionality: file_format, file_metadata, or so

metasyn/filehandler.py
```diff
@@ -0,0 +1,191 @@
...
+            "file_name": self.file_name,
+        }
+
+    def save(self, handler_file):
+        with open(handler_file, "w", encoding="utf-8") as handle:
+            json.dump(self.to_dict(), handle, indent=4)
+
+    @abstractmethod
+    def _write_synthetic(self, df, fp):
+        ...
```
vankesteren, line 42: raise notimplementederror?

metasyn/filehandler.py
vankesteren: as per my main comment

metasyn/filehandler.py
```diff
@@ -0,0 +1,191 @@
...
+
+    @classmethod
+    def _get_df_metadata(cls, fp):
+        try:
+            import pyreadstat
+        except ImportError as err:
+            raise ImportError(
+                "Please install pyreadstat to use the .sav/.zsav file handler.") from err
+
+        pandas_df, prs_metadata = pyreadstat.read_sav(fp, apply_value_formats=True)
```
vankesteren, line 72: metadata can be read using metadata_only=True (see here https://github.com/Roche/pyreadstat?tab=readme-ov-file#reading-only-the-headers). I think this will return an empty dataframe, so 
```suggestion
        _, prs_metadata = pyreadstat.read_sav(fp, metadataonly=True)
```

metasyn/filehandler.py
```diff
@@ -0,0 +1,191 @@
...
+        df = pl.DataFrame(pandas_df)
+        for col in df.columns:
+            col_format = prs_metadata.original_variable_types[col]
+            if (col_format.startswith("F") and col_format.endswith(".0")
+                    and df[col].dtype == pl.Float64):
+                df = df.with_columns(pl.col(col).cast(pl.Int64))
+        return df, prs_metadata
```
vankesteren, line 79: Not sure how this part will work with my main suggestion

metasyn/filehandler.py
```diff
@@ -0,0 +1,191 @@
...
+
+    @classmethod
+    def from_file(cls, fp):
+        if Path(fp).suffix not in [".sav", ".zsav"]:
+            warnings.warn(f"Trying to read file '{fp}' with extension different from .sav or .zsav")
+        if Path(fp).suffix == ".zsav":
+            compress = True
+        else:
+            compress = False
+        df, prs_metadata = cls._get_df_metadata(fp)
```
vankesteren, line 89: prs_metadata -> file_format, or fmt, or so to indicate this is only about IO / file formatting

metasyn/filehandler.py
```diff
@@ -0,0 +1,191 @@
...
+                           file_name=file_format_dict["file_name"])
+    raise ValueError(f"Cannot find file handler with name '{handler_name}'.")
+
+def get_file_handler(fp):
+    suffix = Path(fp).suffix
+
+    for handler_name, handler in _AVAILABLE_FILE_HANDLERS.items():
+        if suffix in handler.extensions:
+            return handler.from_file(fp)
+    raise ValueError(f"Cannot find handler for files with extension '{suffix}'.")
```
vankesteren, line 182: Something like

```suggestion
    raise ValueError(f"No supported file format for files of type '{suffix}'.")
```

metasyn/filehandler.py
```diff
@@ -0,0 +1,191 @@
...
+    def _write_synthetic(self, df, out_fp):
+        df.write_csv(out_fp, **self.metadata)
+
+
+def file_handler_from_dict(file_format_dict):
+    for handler_name, handler in _AVAILABLE_FILE_HANDLERS.items():
+        if file_format_dict["file_handler_name"] == handler_name:
+            return handler(metadata=file_format_dict["format_metadata"],
+                           file_name=file_format_dict["file_name"])
+    raise ValueError(f"Cannot find file handler with name '{handler_name}'.")
```
vankesteren, line 174: Same as below: remove "file handler" from user-facing stuff

metasyn/filehandler.py
vankesteren: this, but better :)

metasyn/metaframe.py
```diff
@@ -473,6 +488,18 @@ def synthesize(self, n: Optional[int] = None, seed: Optional[int] = None) -> pl.
...
 
+    def write_synthetic(self, file_name: Union[None, Path, str] = None,
+                        n: Optional[int] = None, seed: Optional[int] = None,
+                        file_format: Union[None, dict, BaseFileHandler] = None):
+        syn_df = self.synthesize(n, seed)
+        if file_format is not None:
+            self.file_format = file_format
+        if self.file_format is None:
+            raise ValueError("Cannot write synthetic dataset without file handler."
+                             " Use write_synthetic(..., file_handler=your_file_handler)")
```
vankesteren, line 499: Check for file format before synthesizing

```suggestion
        if file_format is not None:
            self.file_format = file_format
        if self.file_format is None:
            raise ValueError("Cannot write synthetic dataset without file handler."
                             " Use write_synthetic(..., file_handler=your_file_handler)")
        syn_df = self.synthesize(n, seed)
```

vankesteren (changes requested):

vankesteren:
Approximate wish of newly discussed UX

```py
import polars as pl
from metasyn import MetaFrame, demo_file
from metasyn import read_csv # thin wrapper around pl.read_csv, passing kwargs
fp = demo_file("spaceship")

# read data and create mf as always
schema = {
    "HomePlanet": pl.Categorical,
    "CryoSleep": pl.Categorical,
    "VIP": pl.Categorical,
    "Destination": pl.Categorical,
    "Transported": pl.Categorical,
}

# read functions give both a df and a fmt
df, fmt = read_csv(fp, try_parse_dates=True, schema_overrides = schema)
mf = MetaFrame.fit_dataframe(df, file_format=fmt)

# this should error before synthesizing if there is no file format (it currently does not)
mf.write_synthetic("test.sav")
```

metasyn/filehandler.py
qubixes: Done.

vankesteren:
to do:
- update GMF schema (make file_format optional)
- only use dicts in api + decorator for custom FileReaders
- warning on write_synthetic with different extension as original file
- file_reader -> file_format in write_synthetic
- to_dict from BaseFileReader should give warning when the decorator `@filereader` is not used
- dev docs for creating a new file reader:
```py
# MWE of custom file reader

from metasyn.filereader import filereader, BaseFileReader

@filereader
class MyFileReader(BaseFileReader):
    name = "SuperSecureFilereader"
    def from_file():
        pass
    def _write_synthetic(self, df: pl.DataFrame, fp: Path):
        print("Sike! I'm not doing anything!")
        pass
```

docs/source/quick_start.rst
```diff
@@ -42,14 +43,23 @@ with the column names that we want to specify as categorical. For this example,
...
+For reading a CSV file, you should give the ``data_types`` dictionary as an argument. 
 
 .. code-block:: python
 
-   df = pl.read_csv(csv_path, schema_overrides=data_types)
+   df, file_reader = ms.read_csv(csv_path, schema_overrides=data_types)
 
 
-This converts the CSV file into a DataFrame named ``df``.
+This converts the CSV file into a DataFrame named ``df``. Additionally, we have a ``file_reader`` object
```
vankesteren, line 61: this will change to file_format dictionary

qubixes: Done

qubixes:
I think the PR should be mergeable now. Test coverage is basically maximized.

PR #356, opened 2025-01-21
Add a parameter to skip checks for empty series

qubixes:
This will not be a functional change in metasyn, or the testing of the core package itself. However, this is needed to fix the testing of the metasyn-disclosure control sub package.


vankesteren (approved):

vankesteren:
this is used here: https://github.com/sodascience/metasyn-disclosure-control/pull/46

PR #368, opened 2025-04-29
Add Stata reader

### LINKED ISSUE #361: Add .dta file reader

qubixes:
This also changes the names of the demo dataset, since periods are not allowed as variable names.

Fixes #361 

- [x] Add writing of categorical with default writer. Not planned for now.
- [x] Add test for types.
- [x] Add mapping for data types.


vankesteren (approved):
Nice. General question: how are categorical columns handled in readstat? Is this nicely mapped to pl.Categorical?

qubixes:
> Nice. General question: how are categorical columns handled in readstat? Is this nicely mapped to pl.Categorical?

Not nicely mapped, but with a correct input file, this mapping should happen I think? (but I don't have a file to check it with). With the default reader, this mapping is currently broken, same as the sav reader.

PR #369, opened 2025-04-29
Improve performance of synthesization

qubixes:
While we're getting worried about synthesizing and how long that takes, we might as well implement vectorization for the Scipy distributions.

Takes the test dataset with 10K rows from ~20 seconds to ~8 seconds.

TODO:

- [x] Refactor base class to add `draw_series`, + raise NotImplementedError if not implemented.
- [x] Improve performance of adding None's
- [x] Tests
- [x] Developer documentation update

### DIFF
--- a/docs/source/developer/distributions.rst
+++ b/docs/source/developer/distributions.rst
@@ -60,7 +60,9 @@ BaseDistribution class has a series of abstract methods that *must be* implemete
 
 If the distribution has subsequently draws that are not independent, it is recommended to implement :meth:`~metasyn.distribution.base.BaseDistribution.draw_reset`. As the name suggests, this method is intended to reset the distribution's drawing mechanism.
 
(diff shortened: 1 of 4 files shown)


vankesteren (commented):
Maybe we should then add `draw_series()` as an optional method in BaseDistribution (the ABC). Is it possible to check instead of hasattr(class, "draw_series") for if class.draw_series() raises NotImplementedError? Or is there a nicer way to _suggest_ to implementers of distributions that they should create a draw_series method if possible?

qubixes:
> Maybe we should then add `draw_series()` as an optional method in `BaseDistribution` (or however the ABC is called 😃) Is it possible to check instead of `hasattr(class, "draw_series")` for `if class.draw_series() raises NotImplementedError`? Or is there a nicer way to _suggest_ to implementers of distributions that they should create a `draw_series()` method if possible?

Yes, that sounds like a better approach. The check would be a `try`/`except` block. This was just a hack to see if you would agree it's a good idea!

qubixes:
@vankesteren I have updated the branch. Shall we merge?

vankesteren (approved):

vankesteren:
Yes! merging...

PR #372, opened 2025-05-13
Add option to set encoding

### LINKED ISSUE #364: add encoding as parameter to csv file reader / writer

qubixes:
This is a bit more difficult than I would have liked, but polars doesn't support writing to anything other than "utf-8".

Fixes #364


vankesteren (approved):

vankesteren:
omg this looks much more difficult indeed. I think this is fine for now, but let's keep an eye on whether polars supports this in the future. Essentially, we're writing the bytes, decoding them, and writing again now. Seems inefficient but also probably not that much of a problem. Feel free to merge!

PR #379, opened 2025-07-08
Add distribution module to exports for type completion

vankesteren:
We often use `metasyn.distribution` when we actually create synthetic data. This change enables type completion and avoids `AttributeError`s from type checkers when doing the following:

```py
import metasyn as ms
ms.distribution.RegexDistribution("[0-5]{10}").draw()
```

It's a small improvement in quality of life :)


qubixes (approved):
If it helps with autocompletion, I'm all for it!

PR #388, opened 2025-09-18
Refactor provider

qubixes:


vankesteren (commented):
Nice, I think the code has become much clearer; some questions remain about renaming provider -> registry. 

Also, DistributionRegistry also contains Fitters. Is that logical? Let's discuss in a bit.

PR #391, opened 2025-10-06
Add citation section to readme

### LINKED ISSUE #386: Add easy-to-copy bibtex citation to readme

vankesteren:
fixes #386


qubixes (approved):
Looks good. One thing to consider perhaps is that not the CITATION.cff is different from the README, which might be confusing. Should we transfer the CITATION.cff to the JOSS paper as well?

PR #406, opened 2025-10-14
Improve the provenance information in the GMF

### LINKED ISSUE #385: Add provenance info to creation method
In the GMF, the creation_method item could integrate provenance information such as which plugin/fitter was used to fit the distribution.

qubixes:
Fixes #385


qubixes:
@vankesteren Done!

vankesteren:
The Titanic one seems to still be the old type but with only the version updated?

PR #414, opened 2026-01-30
Remove some sources of None

qubixes:
None values cause trouble with TOML files, since the TOML language doesn't recognize anything like a none or null value. Should remain backward compatibility with GMF files.

### DIFF
--- a/metasyn/file.py
+++ b/metasyn/file.py
@@ -327,10 +327,13 @@ class SavFileInterface(ReadStatInterface):
             "variable_format": prs_metadata.original_variable_types,
             "compress": compress,
             "variable_display_width": prs_metadata.variable_display_width,
-            "file_label": prs_metadata.file_label,
             "variable_value_labels": prs_metadata.variable_value_labels,
             "variable_measure": prs_metadata.variable_measure,
         }
+        # Workaround for TOML files that don't like None values.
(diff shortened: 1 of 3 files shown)


metasyn/file.py
```diff
@@ -327,10 +327,13 @@ def _extract_metadata(cls, prs_metadata, fp):
...
             "compress": compress,
             "variable_display_width": prs_metadata.variable_display_width,
-            "file_label": prs_metadata.file_label,
             "variable_value_labels": prs_metadata.variable_value_labels,
             "variable_measure": prs_metadata.variable_measure,
         }
+        # Workaround for TOML files that don't like None values.
+        if prs_metadata.file_label is not None:
+            metadata["file_label"] = prs_metadata.file_label
+
```
vankesteren, line 336: Ah ok, so now we just do not include this element in the metadata if it is `None` in prs_metadata

vankesteren (approved):
Nice, approved. I wonder whether we will run into this more often and we should just have the general policy of not including things in our serialization or metadata if they are `None`. Is there an option to create a test for this or am I thinking too big here?

metasyn/file.py
qubixes: Yes, that is the easiest hack (also proposed by the TOML devs) I think, do you have another idea?

PR #421, opened 2026-03-12
docs: add how metasyn works section

zamadye:
hey! i added the "How does metasyn work?" section to explain the process in accessible language.

it covers:
- what metasyn does with the dataframe (column-by-column processing)
- the 5-step process for each column:
  1. look at data type and privacy requirements
  2. select candidate distributions
  3. fit distributions (with privacy constraints if specified)
  4. select best-fitting distribution
  5. store only parameters (not the original data)

i also updated the paper link to point to the published JOSS version (https://doi.org/10.21105/joss.06176) instead of the PDF on github.

let me know if you want any changes to the wording or structure! 👍


vankesteren:
@zamadye please follow the contributing guidelines and disclose any AI use.

PR #423, opened 2026-03-24
Full api reference

### LINKED ISSUE #404: Add full API reference to documentation
It would be nice to have a fully clickable, complete API reference in our documentation so we never have to go through source code to figure out what all our classes and functions do.

I tried to do this as part of #403 but this turned out to be too fickle with our existing automatic documentation.

See also #396

### LINKED ISSUE #409: pyproject.toml update license
follow this new standard: https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license

MSVermet:
1. Added the automatically generated api reference for all modules, functions, and classes to the developer reference section under API reference (issue #404). Left the other API reference sections untouched. 

Is it useful to keep the other API reference sections as well? Or shall I remove those and only have the full api reference there? 

2. And updated the license part in the toml (issue #409)


MSVermet:
There seems to be a problem with setuptools version and the new standard format for license in the toml. Currently the toml has `requires = ["setuptools==75.8.0"...`. If I change this to setuptools>=75 it is not failing anymore. But I'm not sure if this would break anything.

qubixes (changes requested):
That's really nice, thanks! The structure is a little too hierarchical though, especially the demo module, which is quite nested. Also it shouldn't start at the `metasyn` level, but one higher. Also, if you could find how to remove the duplicate reference errors, that would be nice!

MSVermet:
The duplicate description errors are fixed by adding the :no-index: option ([a71d8dc](https://github.com/sodascience/metasyn/pull/423/commits/a71d8dc25ac9cea75c6ea5243528dfda3bda7c7d)). As a result, the members of those modules no longer show up in the navigation menu. I think this is not a problem because all members of those modules are already shown on the page. 

I was able to move the starting level one step higher than `metasyn`. However, this means that modules need to be updated manually whenever they are added, removed, or renamed.

I did not manage to make it less nested. I tried a few things with templates, but these resulted in more duplicate description errors. I’ll keep tinkering with it a bit more.

qubixes:
> The duplicate description errors are fixed by adding the :no-index: option ([a71d8dc](https://github.com/sodascience/metasyn/pull/423/commits/a71d8dc25ac9cea75c6ea5243528dfda3bda7c7d)). As a result, the members of those modules no longer show up in the navigation menu. I think this is not a problem because all members of those modules are already shown on the page.
> 
> I was able to move the starting level one step higher than `metasyn`. However, this means that modules need to be updated manually whenever they are added, removed, or renamed.
> 
> I did not manage to make it less nested. I tried a few things with templates, but these resulted in more duplicate description errors. I’ll keep tinkering with it a bit more.

Thanks again for looking into the issue.  The problem with the duplicate index is that the links to the code elements (classes, functions) doesn't work anymore because it doesn't know which one to link to.

If it turns out too difficult, we can simply revert back to the more manual way of doing things, while having some improvements. Either way, this has been an unresolved issue for a while, and it was unresolved for a reason.

MSVermet:
I've made it a little less nested without duplication errors. The modules now also show the documentation of their classes/functions/methods on the same page. Only if it contains submodules like `distribution` for example it is still nested, but I think this is a cleaner way of showing it rather than having all the contents of all the submodules on one page.

qubixes (approved):
Thanks a lot! I think it is nicer now, and hopefully less maintenance intensive.

PR #427, opened 2026-04-17
Improve the performance of read/write stat

qubixes:
I have done some performance tests and this is way faster than the method before. For big dataset reading sav/dta files should be many times faster when using sampling of rows.


MSVermet (approved):
Great, this fixes the issue I sent you!

PR #428, opened 2026-04-21
Fix truncated normal

### LINKED ISSUE #401: Remove tolerance for minimum/maximum of truncated distributions?
From @MWeiper: 
> I have data that starts at 0, but the truncated normal distribution shows a minimum of `-1e-08`. So there is a chance (however small) of generating negative values. This is strange to me.

Should we remove this tolerance?

qubixes:
Fixes #401

### DIFF
--- a/metasyn/distribution/normal.py
+++ b/metasyn/distribution/normal.py
@@ -153,7 +153,7 @@ class ContinuousTruncatedNormalFitter(BaseFitter):
         upper = series.max()
         if lower == upper:
             return self.distribution(lower-1e-8, upper+1e-8, lower, 1)
-        return self._fit_with_bounds(series, series.min(), series.max())
+        return self._fit_with_bounds(series, lower, upper)
 
     def _fit_with_bounds(self, values, lower, upper):
         def minimizer(param):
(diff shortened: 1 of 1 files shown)


metasyn/distribution/normal.py
```diff
@@ -149,9 +149,11 @@ class ContinuousTruncatedNormalFitter(BaseFitter):
...
 
     def _fit(self, series):
-        lower = series.min() - 1e-8
-        upper = series.max() + 1e-8
-        return self._fit_with_bounds(series, lower, upper)
+        lower = series.min()
+        upper = series.max()
+        if lower == upper:
+            return self.distribution(lower-1e-8, upper+1e-8, lower, 1)
+        return self._fit_with_bounds(series, series.min(), series.max())
```
maartenschermer: `min()` and `max()` are probably not very calculation heavy, but since `lower` and `upper` are already there, it makes sense to do `return self._fit_with_bounds(series, lower, upper)` here.

qubixes: Thanks!

maartenschermer (approved):

PR #442, opened 2026-06-25
metaframe-builder: add arithmetic operations and logical conditions

MSVermet:
Implemented logical conditions: greater than, less than, and not equal.

In the notebook (exec_builder) you can see an example of how it works. Is this how it supposed to work?

### DIFF
--- a/examples/exec_builder.ipynb
+++ b/examples/exec_builder.ipynb
@@ -2,7 +2,7 @@
  "cells": [
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": null,
    "id": "dd6c6c6c",
    "metadata": {},
    "outputs": [],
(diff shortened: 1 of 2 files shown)


qubixes:
> Implemented logical conditions: greater than, less than, and not equal.
> 
> In the notebook (exec_builder) you can see an example of how it works. Is this how it supposed to work?

Thanks! Yes, but you can even do 

```python
builder["Grownup"].distribution = ColumnReference("Age") > 18
```
There are a few more operators that might be interesting for our usecase, see: https://www.pythonmorsels.com/every-dunder-method/. For example not equal, greater equal, less equal, pow. Perhaps you can implement the ones that seem relevant to you? ```__and__``` and ```__or__``` might also be very useful.

MSVermet:
> > Implemented logical conditions: greater than, less than, and not equal.
> > In the notebook (exec_builder) you can see an example of how it works. Is this how it supposed to work?
> 
> Thanks! Yes, but you can even do
> 
> ```python
> builder["Grownup"].distribution = ColumnReference("Age") > 18
> ```
> 
> There are a few more operators that might be interesting for our usecase, see: https://www.pythonmorsels.com/every-dunder-method/. For example not equal, greater equal, less equal, pow. Perhaps you can implement the ones that seem relevant to you? `__and__` and `__or__` might also be very useful.

So if these logical conditions are not meant to do much more than just applying the operator like you did in the example, why should we implement them?

qubixes:
> > > Implemented logical conditions: greater than, less than, and not equal.
> > > In the notebook (exec_builder) you can see an example of how it works. Is this how it supposed to work?
> > 
> > 
> > Thanks! Yes, but you can even do
> > ```python
> > builder["Grownup"].distribution = ColumnReference("Age") > 18
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > There are a few more operators that might be interesting for our usecase, see: https://www.pythonmorsels.com/every-dunder-method/. For example not equal, greater equal, less equal, pow. Perhaps you can implement the ones that seem relevant to you? `__and__` and `__or__` might also be very useful.
> 
> So if these logical conditions are not meant to do much more than just applying the operator like you did in the example, why should we implement them?

Because it is more legible. Without the dunder you can still implement the operators if you want of course, but `a&(b|c)` is much more legible than `AndOperator(a, OrOperator(b, c))`.

MSVermet:
> > > > Implemented logical conditions: greater than, less than, and not equal.
> > > > In the notebook (exec_builder) you can see an example of how it works. Is this how it supposed to work?
> > > 
> > > 
> > > Thanks! Yes, but you can even do
> > > ```python
> > > builder["Grownup"].distribution = ColumnReference("Age") > 18
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > There are a few more operators that might be interesting for our usecase, see: https://www.pythonmorsels.com/every-dunder-method/. For example not equal, greater equal, less equal, pow. Perhaps you can implement the ones that seem relevant to you? `__and__` and `__or__` might also be very useful.
> > 
> > 
> > So if these logical conditions are not meant to do much more than just applying the operator like you did in the example, why should we implement them?
> 
> Because it is more legible. Without the dunder you can still implement the operators if you want of course, but `a&(b|c)` is much more legible than `AndOperator(a, OrOperator(b, c))`.

Never mind. I didn't fully understand how dunders worked, now I do. 

I went through the list of dunder methods and I think you already named the relevant ones. We can always add other methods later. I still need to check if `__and__` and `__or__` are working, I ran into an error but not sure if it had to do with the bad example I applied it on. I'll check this later.

qubixes:
The only one remaining I can think of would be the `__invert__` dunder for booleans. Then later a lot of tests need to be written of course.

MSVermet:
Applying the invert like below, does not work. 

`builder["not grownup"].distribution = not ColumnReference("Grownup")`

TypeError: Unknown type for recipe: <class 'bool'>.

But this works: `ColumnReference("Grownup") == False`

qubixes (approved):
Thanks!

PR #453, opened 2026-07-17
Add fit log

qubixes:
The functionality of the fit log is meant to audit the process of fitting the data. You can see which categories got deleted while following the disclosure control process. You can also see which distributions are considered during the fitting process. And for the micro-aggregation process, you can see which parameters the fitting process eventually landed on.

The API change is rather a big one, and I wish it weren't necessary, but I think the change will allow more flexibility in the future.

This also removes the toml GMF file, since most of that has now been superseeded by the fit log.

### DIFF
--- a/examples/getting_started.ipynb
+++ b/examples/getting_started.ipynb
@@ -306,7 +306,7 @@
      "name": "stderr",
      "output_type": "stream",
      "text": [
-      "100%|████████████████████████████████████████████████████████████| 13/13 [00:04<00:00,  2.90it/s]\n"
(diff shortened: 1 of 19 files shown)


MSVermet:
How/where is a user supposed to see this log?

qubixes:
> How/where is a user supposed to see this log?

It is an attribute of the MetaFrameBuilder object, called `fit_log`.

```python
print(builder.fit_log)
print(builder.fit_log["some_column"])
builder.fit_log.save_md("fitlog.md")
builder.fit_log.save_csv("fitlog.csv")
```
I have added an extra section to the tutorial that explains this a bit and shows the two exporting options.

MSVermet (approved):

PR #456, opened 2026-08-19
Update maintainers, developers, etc.

### LINKED ISSUE #441: Update maintainers on readthedocs
Erik-Jan and Samuel are not maintaining metasyn anymore.

qubixes:
I have updated the current list of maintainers, developers. It would be especially nice to get feedback on how (former) developers would like to be mentioned. The current proposal is to put the significant developers into the hall of fame section of the about section in the readthedocs documentation, but any suggestions are welcome! Ideally, everyone approves / suggests a new way of being cited:

- [x] @vankesteren 
- [ ] @Samuwhale 
- [x] @maartenschermer 
- [x] @MSVermet 

Fixes #441

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,7 +6,6 @@ build-backend = "setuptools.build_meta"
 name = "metasyn"
 authors = [
     { name = "Raoul Schram", email = "r.d.schram@uu.nl" },
-    { name = "Erik-Jan van Kesteren", email = "e.vankesteren1@uu.nl" },
 ]
 description = "Package for creating synthetic datasets while preserving privacy."
 readme = "README.md"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,7 +9,9 @@ authors = [
     { name = "Raoul Schram", email = "r.d.schram@uu.nl" }
 ]
 maintainers = [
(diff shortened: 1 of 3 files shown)


pyproject.toml
vankesteren, line 1: I am ok with being removed from the toml, but a little discussion is warranted maybe. The [official spec](https://packaging.python.org/en/latest/specifications/pyproject-toml/#pyproject-toml-spec) States the following:

> The people or organizations considered to be the “authors” of the project. The exact meaning is open to interpretation — it may list the original or primary authors, current maintainers, or owners of the package.

There is also the "maintainer" field. Maybe Raoul could appear in both and people who have done a substantial contribution could appear in the authors? Then what is a "substantial contribution" is up for debate of course.

R solves this better, there you can specify a role for each author.

Again, don't let my opinions block you, just wanted to start the discussion!

qubixes: @vankesteren Thanks for looking it up! Given that reading, I definitely think that you should be in the authors section, but I have removed the email address for now. Let me know if you want another email address added. It would be nice if the spec was a little more precise, but I suppose it also allows every project to have their own policy.

Just to emphasize, it was just a suggestion, so if any of you want to mentioned somewhere else (or in the toml), I'm all for it!

maartenschermer (approved):

PR #461, opened 2026-08-20
update tutorials; fixed new branch

MSVermet:
Created a new fork and a new branch. I'm not allowed to remove the old branch 'update-tutorials', can you remove it?

Could you review the tutorial?


examples/3_additional_features.ipynb
```diff
@@ -0,0 +1,969 @@
...
+    "    IfThenElse,\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "sec1_title",
+   "metadata": {},
+   "source": [
+    "### 1. Synthesising time-series data\n",
```
qubixes: We have generally been using American spelling, so for consistency, let's use `synthesizing`.

examples/3_additional_features.ipynb
```diff
@@ -0,0 +1,969 @@
...
+       "  white-space: pre-wrap;\n",
+       "}\n",
+       "</style>\n",
+       "<small>shape: (5, 8)</small><table border=\"1\" class=\"dataframe\"><thead><tr><th>Patient_id</th><th>Admission_date</th><th>Discharge_date</th><th>Length_of_stay_days</th><th>Age</th><th>Height_cm</th><th>Weight_kg</th><th>Sex</th></tr><tr><td>i64</td><td>datetime[μs]</td><td>datetime[μs]</td><td>i64</td><td>i64</td><td>i64</td><td>i64</td><td>str</td></tr></thead><tbody><tr><td>1</td><td>2023-01-04 00:00:00</td><td>2023-01-06 00:00:00</td><td>2</td><td>44</td><td>164</td><td>67</td><td>&quot;F&quot;</td></tr><tr><td>2</td><td>2023-01-08 00:00:00</td><td>2023-01-18 00:00:00</td><td>10</td><td>78</td><td>154</td><td>57</td><td>&quot;F&quot;</td></tr><tr><td>3</td><td>2023-01-08 00:00:00</td><td>2023-01-19 00:00:00</td><td>11</td><td>18</td><td>167</td><td>77</td><td>&quot;M&quot;</td></tr><tr><td>4</td><td>2023-01-08 00:00:00</td><td>2023-01-24 00:00:00</td><td>16</td><td>83</td><td>177</td><td>95</td><td>&quot;F&quot;</td></tr><tr><td>5</td><td>2023-01-15 00:00:00</td><td>2023-01-21 00:00:00</td><td>6</td><td>9</td><td>138</td><td>25</td><td>&quot;M&quot;</td></tr></tbody></table></div>"
```
qubixes: The 'Sex' column should be a categorical column.

examples/3_additional_features.ipynb
qubixes, line 1: The example with the length of stays is in principle nice. However, the datatype at the end is not the same as the input type of the dataframe. So in the end I'm worried about it not being such a great example after all. What happens if you try to convert it back to an integer? Otherwise, you can try to read it in as a duration instead.

metasyn/demo/dataset.py
```diff
@@ -199,6 +199,23 @@ def name(self):
     @property
     def schema(self):
         return {}
+    
+@register
```
qubixes, line 204: Nice! The Admission_date and Discharge_date are datetimes, could you fix them to dates?

examples/3_additional_features.ipynb
qubixes, line 1: I was looking at the `height_cm` and `weight_kg`, and it would be nice perhaps if we define the height as a function as a function of the age, and the weight as a function of the height. What do you think?

qubixes (changes requested):
Thanks for the update. I have some additional suggestions for improvement. I think we're almost there!

qubixes (commented):
Also, it would be nice if in this branch the tests were fixed.

examples/3_additional_features.ipynb
MSVermet: Reading it in as duration type didn't work, but converting back to integer works.

MSVermet: Good idea! That's a nicer example than applying it as a function of age twice.

qubixes (approved):
Thanks for fixing the tests!!

